PR #6, opened 2023-02-10
Rework assign operations

mingweishih:
Details of the PR
- Support "var = value", "value = var", and "var = var"
- Fall back the '=' to comparison when both operators are defined
- Correctly support variable shadowing with "var := value"

Also, remove the unused Variable struct

Signed-off-by: Ming-Wei Shih <mishih@microsoft.com>


anakrish (approved):
LGTM

PR #18, opened 2023-05-14
support of or-functions

eric-therond:
Probably not a clean code, I am learning both rust and rego languages ...

I tried to add support of mutiple definitions of a function

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -1259,6 +1259,7 @@ impl<'source> Interpreter<'source> {
         let fcns = fcns_rules.clone();
 
         let mut results: Vec<Value> = Vec::new();
+        let mut errors: Vec<anyhow::Error> = Vec::new();
         for fcn_rule in fcns {
             let (args, output_expr, bodies) = match fcn_rule {
                 Rule::Spec {
@@ -1304,7 +1305,15 @@ impl<'source> Interpreter<'source> {
 
             // Set the arguments scope.
             self.scopes.push(args_scope);
-            let value = self.eval_rule_bodies(ctx, span, bodies)?;
+            let value = match self.eval_rule_bodies(ctx, span, bodies) {
+                Ok(v) => v,
+                Err(e) => {
+                    // If the rule produces an error, save the error.
+                    errors.push(e);
+                    continue;
+                }
+            };
+
             let result = match &value {
                 Value::Set(s) if s.len() == 1 => s.iter().next().unwrap().clone(),
                 Value::Set(s) if !s.is_empty() => {
@@ -1331,7 +1340,11 @@ impl<'source> Interpreter<'source> {
         }
 
         if results.is_empty() {
-            return Ok(Value::Undefined);
+            if errors.is_empty() {
+                return Ok(Value::Undefined);
+            } else {
+                return Err(anyhow!(errors[0].to_string()));
+            }
         }
 
         // all defined values should be the equal to the same value that should be returned
@@ -1342,7 +1355,6 @@ impl<'source> Interpreter<'source> {
                 "functions must not produce multiple outputs for same inputs",
             ));
         }
-
         Ok(results[0].clone())
     }
 
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -26,7 +26,7 @@ pub struct Interpreter<'source> {
     // TODO: handle recursive calls where same expr could have different values.
     loop_var_values: BTreeMap<&'source Expr<'source>, Value>,
     contexts: Vec<Context<'source>>,
-    functions: HashMap<String, &'source Rule<'source>>,
+    functions: HashMap<String, Vec<&'source Rule<'source>>>,
     rules: HashMap<String, Vec<&'source Rule<'source>>>,
     default_rules: HashMap<String, Vec<(&'source Rule<'source>, Option<String>)>>,
     processed: BTreeSet<&'source Rule<'source>>,
@@ -1180,7 +1180,7 @@ impl<'source> Interpreter<'source> {
         }
     }
 
-    fn lookup_function(&self, fcn: &'source Expr<'source>) -> Result<&'source Rule<'source>> {
+    fn lookup_function(&self, fcn: &'source Expr<'source>) -> Result<&Vec<&'source Rule<'source>>> {
         let mut path = Self::get_path_string(fcn, None)?;
         if !path.starts_with("data.") {
             path = self.current_module_path.clone() + "." + &path;
@@ -1239,7 +1239,7 @@ impl<'source> Interpreter<'source> {
         fcn: &'source Expr<'source>,
         params: &'source Vec<Expr<'source>>,
     ) -> Result<Value> {
-        let fcn_rule = match self.lookup_function(fcn) {
+        let fcns_rules = match self.lookup_function(fcn) {
             Ok(r) => r,
             _ => {
                 // Look up builtin function.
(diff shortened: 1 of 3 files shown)


anakrish:
@eric-therond Thanks for the contribution. Can you sign off you commit?

anakrish:
The PR looks good to me. Are there other tests that you can think of?

eric-therond:
@anakrish I signed-off the commit
I'am ok with the tests for now

src/interpreter.rs
```diff
@@ -1256,69 +1256,99 @@ impl<'source> Interpreter<'source> {
...
+        for fcn_rule in fcns_rule {
+            fcns.push(fcn_rule);
         }
 
-        let mut args_scope = Scope::new();
-        for (idx, a) in args.iter().enumerate() {
-            let a = match a {
-                Expr::Var(s) => s.text(),
-                _ => unimplemented!("destructuring function arguments"),
+        let mut results: Vec<Result<Value>> = Vec::new();
```
mingweishih: Can the type just be `Vec<Value>`?

src/interpreter.rs
```diff
@@ -1256,69 +1256,99 @@ impl<'source> Interpreter<'source> {
...
+            let scopes = std::mem::take(&mut self.scopes);
 
-        // Restore local variables for current context.
-        self.scopes = scopes;
-        result
+            // Set the arguments scope.
+            self.scopes.push(args_scope);
+            let value = self.eval_rule_bodies(ctx, span, bodies)?;
+            let result = match &value {
+                Value::Set(s) if s.len() == 1 => Ok(s.iter().next().unwrap().clone()),
```
mingweishih: If the results is `Vec<Value>`, you can remove `Ok(...)` wrapping here and below.

src/interpreter.rs
```diff
@@ -1256,69 +1256,99 @@ impl<'source> Interpreter<'source> {
...
+
+            // If the function successfully executed and returned true, not necessary to execute the others
+            if let Ok(Value::Bool(true)) = result {
+                return result;
+            }
+
+            results.push(result);
+        }
+
+        if let Ok(value0) = &results[0] {
```
mingweishih: Assuming results is `Vec<Value>`, you could do something like:
```rust
        if results.is_empty() {
            bail!("internal error");
        }

        if results.iter().any(|value| value != results[0]) {
            return Err(span.source.error(
                span.line,
                span.col,
                "functions must not produce multiple outputs for same inputs",
            ));
        }

        Ok(results[0])
```

anakrish (approved):
LGTM.

Thanks for the contributtion.

anakrish:
@eric-therond Kindly address suggestions from @mingweishih

src/interpreter.rs
eric-therond: It makes sense to me, thanks

tests/interpreter/cases/call/or.yaml
```diff
@@ -0,0 +1,59 @@
...
+    query: data.test
+    error: functions must not produce multiple outputs for same inputs
+
+  - note: or-ok
+    data: {}
+    modules:
+      - |
+        package test
+
+        inc(x) = x + 1
```
anakrish, line 24: We also need to add another test case for cases where one of the function definitions evaluates to `undefined`:
```
        inc(x) = x + 1
        inc(x) = x + 1
        inc(x) = y {
           x > 10 # This will evaluate to false.
           y = 100 # y will be undefined.
        }

        a1 = inc(5)
```

The logic needs to be fixed to handle this. Undefined definitions are ignored.

tests/interpreter/cases/call/or.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - |
+        package test
+
+        inc(x) = 1 == x
+        inc(x) = 2 == x
+
+        a1 = inc(2)
+    query: data.test
+    want_result:
+      a1: true
```
anakrish, line 44: The above example produces `eval_conflict_error: functions must not produce multiple outputs for same inputs` in Rego playground.

src/interpreter.rs
```diff
@@ -1256,69 +1256,99 @@ impl<'source> Interpreter<'source> {
...
+
+                // If the function execution resulted in undefined, then propagate it.
+                Value::Undefined => Value::Undefined,
+                _ => bail!("internal error: function did not return set {value:?}"),
+            };
+
+            // Restore local variables for current context.
+            self.scopes = scopes;
+
+            // If the function successfully executed and returned true, not necessary to execute the other
```
anakrish: This assumption may not be correct. It seems that Rego requires all the definitions to be executed even if one definition returned true.

eric-therond: You're absolutely right, I think I mixed up two different rego objects: "incremental rules" and "multiple function definitions" (I told you I was new to rego !). 

To my understanding, rules with the same name should be executed one after another until true is returned, but for functions, as you said, all functions with the same name should be executed, and the return value should be the same, regardless of the value set (defined as in your example).

I will fix that tomorrow

src/interpreter.rs
```diff
@@ -1239,7 +1239,7 @@ impl<'source> Interpreter<'source> {
         fcn: &'source Expr<'source>,
         params: &'source Vec<Expr<'source>>,
     ) -> Result<Value> {
-        let fcn_rule = match self.lookup_function(fcn) {
+        let fcns_rule = match self.lookup_function(fcn) {
```
anakrish: `fcn_rules` would be a better name since there are many rules for the same function.

src/interpreter.rs
```diff
@@ -1256,69 +1256,97 @@ impl<'source> Interpreter<'source> {
...
-                span.line,
-                span.col,
-                format!(
-                    "mismatch in number of arguments. supplied {}, expected {}",
-                    params.len(),
-                    args.len()
-                )
-                .as_str(),
-            ));
+        let mut fcns: Vec<&Rule> = Vec::new();
```
anakrish: We don't need this copying to `fcns` anymore.

src/interpreter.rs
```diff
@@ -1256,69 +1256,97 @@ impl<'source> Interpreter<'source> {
...
+            results.push(result);
+        }
+
+        if results.is_empty() {
+            bail!("internal error");
+        }
+
+        if results
+            .iter()
+            .any(|value| *value != results[0] && *value != Value::Undefined)
```
anakrish: If the first result is `Undefined`, then this logic won't work.

tests/interpreter/cases/call/or.yaml
```diff
@@ -0,0 +1,77 @@
...
+        inc(x) = y {
+           x > 10 # This will evaluate to false.
+           y = 100 # y will be undefined.
+        }
+
+        a1 = inc(5)
+    query: data.test
+    want_result:
+      a1: 6
+
```
anakrish: Add test cases for the following scenarios:
1, First definition produces undefined value.
2. All definitions produce undefined value.

Confirm expected behavior in Rego playground.

src/interpreter.rs
eric-therond: `fcns_rules` sound ok ?

src/interpreter.rs
eric-therond: I have a borrowing error without the copy that I am not able to fix by myself

tests/interpreter/cases/call/or.yaml
eric-therond: I changed a bit my approach to cover all the cases

src/interpreter.rs
anakrish: Sounds good.

src/interpreter.rs
anakrish: Ok. The compiler thinks that `fcns_rules` could be modified by the subsequent code.

anakrish: You can simplify the code using the clone method.
```suggestion
        let fcns = fcns_rules.clone();
```

src/interpreter.rs
```diff
@@ -1256,69 +1256,105 @@ impl<'source> Interpreter<'source> {
...
+                        span.line,
+                        span.col,
+                        format!("function produced multiple outputs {value:?}").as_str(),
+                    ))
+                }
+                // If the function successfully executed, but did not return any value, then return true.
+                Value::Set(s) if s.is_empty() && output_expr.is_none() => Value::Bool(true),
+
+                // If the function execution resulted in undefined, then propagate it.
+                Value::Undefined => Value::Undefined,
```
anakrish, line 1330: We could have a strategy where we don't gather Undefined values. In that case,

1. If `results` is empty, that means all the rules evaluated to `Undefined`.
2. If `results` is not empty, we need to check that all the values are the same.

src/interpreter.rs
```diff
@@ -1256,69 +1256,105 @@ impl<'source> Interpreter<'source> {
...
-        let ctx = Context {
-            key_expr: None,
-            output_expr,
-            value: Value::new_set(),
-        };
+        match results.len() {
+            0 => bail!("internal error"),
+            1 => Ok(results[0].clone()),
+            _ => {
+                let mut defined_results = results.clone();
```
anakrish: We can simplify the logic by avoiding pushing `Undefined` values altogether.

tests/interpreter/cases/call/or.yaml
```diff
@@ -0,0 +1,114 @@
...
+        inc(x) = x + 1
+        inc(x) = x + 1
+
+        a1 = inc(5)
+    query: data.test
+    want_result:
+      a1: 6
+      
+
+  - note: or-all-undefined
```
anakrish, line 96: Good tests.

anakrish (commented):
Do think about:
1. How we can simplify the logic and possibly avoid extra copying
2. Any more test cases that will lock down the logic.

tests/interpreter/cases/call/or.yaml
```diff
@@ -0,0 +1,114 @@
...
+           y1 = 100 # y will be undefined.
+        }
+        inc(x2) = y2 {
+           x2 > 10 # This will evaluate to false.
+           y2 = 100 # y will be undefined.
+        }
+
+        a1 = inc(5)
+    query: data.test
+    want_result: {}
```
anakrish, line 113: There is one more case to handle:
```
fcn(x) = y {
  y = x + 1
}

fcn(x) = y {
   y = concat(" ", ["hello", x])
}

a1 = fcn("world")
a2 = fcn(5)
```

This produces:
```
{
    "a1": "hello world",
    "a2": 6
}
```

The definitions that produce an error are ignored. If all definitions produce errors, then one of the errors is propagated.

anakrish (commented):
@eric-therond Please squash your commits.

tests/interpreter/cases/call/or.yaml
eric-therond: I added the test case without being able to fix it in this PR. 
I left a comment to not forget about it.

Unless I missed a simple solution, there is a lot of code to rewrite to support that ? especially how errors are handled.

anakrish: For minimal changes, we could slightly tweak your current logic:

1. Also push errors into a vector `errors` within the loop.
2. Then
   - If `results` is not empty then ignore `errors` and return the result after asserting that all the values are same.
   - If `results` is empty, then 
       - If `errors` is not empty, then return the first error.
       - If `errors` is empty, then return Undefined.

src/interpreter.rs
```diff
@@ -1256,69 +1256,94 @@ impl<'source> Interpreter<'source> {
...
+            // Back up local variables of current function and empty
+            // the local variables of callee function.
+            let scopes = std::mem::take(&mut self.scopes);
 
-        // Back up local variables of current function and empty
-        // the local variables of callee function.
-        let scopes = std::mem::take(&mut self.scopes);
+            // Set the arguments scope.
+            self.scopes.push(args_scope);
+            let value = self.eval_rule_bodies(ctx, span, bodies)?;
```
anakrish: You could do something like this to accumulate errors
```suggestion
            let value = match self.eval_rule_bodies(ctx, span, bodies) {
               Ok(v) => v
               Err(e) => {
                    // If the rule produces an error, save the error.
                    errors.push(Err(e));
                    continue;
               }
            };
```

src/interpreter.rs
```diff
@@ -1256,69 +1256,94 @@ impl<'source> Interpreter<'source> {
...
-                span.col,
-                format!(
-                    "mismatch in number of arguments. supplied {}, expected {}",
-                    params.len(),
-                    args.len()
-                )
-                .as_str(),
-            ));
-        }
+        let mut results: Vec<Value> = Vec::new();
```
anakrish, line 1261:
```suggestion
        let mut results = Vec::new();
        let mut errors = Vec::new();
```

src/interpreter.rs
```diff
@@ -1256,69 +1256,94 @@ impl<'source> Interpreter<'source> {
...
+        }
 
-        // Set the arguments scope.
-        self.scopes.push(args_scope);
-        let value = self.eval_rule_bodies(ctx, span, bodies)?;
-        let result = match &value {
-            Value::Set(s) if s.len() == 1 => Ok(s.iter().next().unwrap().clone()),
-            Value::Set(s) if !s.is_empty() => Err(span.source.error(
+        if results.is_empty() {
+            return Ok(Value::Undefined);
```
anakrish:
```suggestion
           if errors.is_empty() {
               return Ok(Value::Undefined);
           } else {
               return errors[0];
           }
```

tests/interpreter/cases/call/or.yaml
eric-therond: I didn't know that errors are propagated like that in Rust ... thanks for the hint, it's super clear.

I still had a problem, the test case `or-one-error` returns
```
{
  "a1": [
    "hello world"
  ],
  "a2": [
    6
  ]
}
```

but in rego playground it is:
```
{
    "a1": "hello world",
    "a2": 6
}
```

I didn't investigate why this difference.

anakrish: Thanks for pointing it out. I will take a look at it.

anakrish:
@eric-therond Can you squash you commits?

I will then merge the PR and then investigate fixing the return values in the case you pointed out.

Thanks for your contribution!

PR #19, opened 2023-06-20
fix bitwise.and and add tests

eric-therond:


src/builtins/bitwise.rs
```diff
@@ -14,7 +14,7 @@ use anyhow::Result;
 
 pub fn register(m: &mut HashMap<&'static str, builtins::BuiltinFcn>) {
     m.insert("bits.and", and);
-    m.insert("bits.and", lsh);
+    m.insert("bits.lsh", lsh);
```
anakrish, line 17: Good catch!

tests/interpreter/cases/builtins/bitwise/tests.yaml
```diff
@@ -0,0 +1,103 @@
...
+# Licensed under the MIT License.
+
+cases:
+  - note: bits-error
+    data: {}
+    modules:
+      - |
+        package test
+
+        x = bits.and(1, "str")
```
anakrish: Also lock down with a test for `and(string, number)` as well as with more than two parameters.

anakrish (commented):
Looks good to me.  Thanks for your contribution!

To make it complete, we would need tests for the following cases:
1) Less number of arguments than expected
2) More number of arguments than expected
3) Wrong type for argument at each position.

If the tests become large, we could split the files to `bitwise/and.yaml`, `bitwise/lsh.yaml` etc.

PR #20, opened 2023-06-22
Ensure that scopes are cleaned up correctly upon error.

anakrish:
When evaluating rules, upon error the last pushed scope wasn't being popped from the stack of scopes. This causes incorrect behavior when there are multiple definitions for the same rule name.

The fix is to make sure that the scopes are popped manually upon encountering errors.

Once the interpreter logic is locked down, then we need to clean up scope management using Drop functions so that the cleanup happens even during short circuited return.


anakrish:
CC: @eric-therond

PR #21, opened 2023-06-26
negation of an undefined value should return true

eric-therond:
When I read this discussion https://github.com/open-policy-agent/opa/issues/1622#issuecomment-520547385
It's not clear for me what should be the value of a negated undefined value

But currently, on rego playground, `not undefined` returns `true`

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -760,7 +760,10 @@ impl<'source> Interpreter<'source> {
                 value,
                 collection,
             } => self.eval_some_in(span, key, value, collection, stmts)?,
-            Literal::NotExpr { expr, .. } => matches!(self.eval_expr(expr)?, Value::Bool(false)),
+            Literal::NotExpr { expr, .. } => {
+                // https://github.com/open-policy-agent/opa/issues/1622#issuecomment-520547385
+                matches!(self.eval_expr(expr)?, Value::Bool(false) | Value::Undefined)
+            }
             Literal::Every {
                 span,
                 key,
--- a/tests/interpreter/cases/negation/tests.yaml
+++ b/tests/interpreter/cases/negation/tests.yaml
@@ -1,0 +1,66 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+cases:
+  - note: not-false
+    data: {}
+    modules:
+      - |
+        package test
+
+        x[a] {
+          not 1 == 2
+          a := "hello"
+        }
+    query: data.test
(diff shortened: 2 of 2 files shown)


anakrish (commented):
LGTM,

Thanks for the contribution.

Can you also add the following test:
```
        package test

        import future.keywords.if

        p if false # undefined value

        q1 {
          not p # `not undefined` evaluates to true
        }
        
        q2 = [p] # q2 = [undefined] = undefined
        
        q3 {
          not [p] # `not undefined` evaluates to true
        }

        q4 {
          not q3 # `not true` making it undefined
        }
```

Rego playground behaves differently, which we might need to follow up about.

eric-therond:
It's done

I found [this issue](https://github.com/open-policy-agent/opa/issues/1877) explaining that queries could be rewritten in REPL and produce unexpected results regarding negated expressions.

Maybe it can explain the difference?

anakrish:
@eric-therond Thanks for finding out why OPA Rego behaves differently.

src/interpreter.rs
```diff
@@ -760,7 +760,9 @@ impl<'source> Interpreter<'source> {
                 value,
                 collection,
             } => self.eval_some_in(span, key, value, collection, stmts)?,
-            Literal::NotExpr { expr, .. } => matches!(self.eval_expr(expr)?, Value::Bool(false)),
+            Literal::NotExpr { expr, .. } => {
```
anakrish, line 763: Add link to the discussion https://github.com/open-policy-agent/opa/issues/1622#issuecomment-520547385 in comments here.

tests/interpreter/cases/negation/tests.yaml
```diff
@@ -0,0 +1,65 @@
...
+        x[a] {
+          not myequal(1)
+          a := "hello"
+        }
+    query: data.test
+    want_result:
+      x: 
+        set!: [hello]
+
+  - note: not-more-undefined
```
anakrish:
```suggestion
  - note: not-more-undefined-OPA-INCOMPATIBLE
```
We could used a naming convention such as above to keep track of behavioral differences from OPA.
Also add the link to https://github.com/open-policy-agent/opa/issues/1877 in comments.

anakrish (commented):
LGTM.

I have couple of suggestions.

tests/interpreter/cases/negation/tests.yaml
eric-therond: Done

PR #22, opened 2023-07-06
test: improve api

eric-therond:
Hello
my first need is to eval the same rule on different inputs in the more optimized way

OPA provides some high-level API to do that (and more):
https://pkg.go.dev/github.com/open-policy-agent/opa@v0.54.0/rego#example-Rego.PrepareForEval

This PR is the closer I was able to do with regorus

Before going further, I want to know if this change is going in the good direction


src/interpreter.rs
```diff
@@ -1924,27 +1924,26 @@ impl<'source> Interpreter<'source> {
                 head: rule_head,
                 bodies: rule_body,
             } => {
-                if matches!(rule_head, RuleHead::Func { .. }) {
-                    return Ok(());
-                }
+                if !matches!(rule_head, RuleHead::Func { .. }) {
```
anakrish, line 1970: Good catch. Previously, the state wasn't being managed cleanly on early exit.

We'd want to move line 1952 into this block so that only non function rules are marked as processed.

src/interpreter.rs
```diff
@@ -1957,29 +1956,42 @@ impl<'source> Interpreter<'source> {
         }
     }
 
-    pub fn eval(
+    pub fn eval_rule_input(
```
anakrish: `eval_rule_with_input` seems like a better name.

tests/interpreter/mod.rs
```diff
@@ -165,6 +165,68 @@ pub fn assert_match(computed: Value, expected: Value) {
...
+        file: "<query.rego>",
+        contents: query,
+        lines: query.split('\n').collect(),
+    };
+    let mut parser = Parser::new(&source)?;
+    let expr = parser.parse_membership_expr()?;
+    let result_eval_snippet = interpreter.eval_query_snippet(&expr, enable_tracing);
+
+    // Now eval the first rule on a specific input
+    if let Some(module) = &modules.get(0) {
```
anakrish: Should we eval the first rule or the query expr?

anakrish (commented):
@eric-therond Thanks for taking this on. You are on the right track. 

When reevaluating an expr with new inputs, we'd want to mark all rules that depended on previous input values as unprocessed.

src/interpreter.rs
anakrish, line 1970: It might make sense for us to use something like [scopeguard](https://github.com/bluss/scopeguard) so that the cleanup code executes even in case of early returns.

tests/interpreter/cases/input/tests.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+cases:
+  - note: input-basic-1
+    data: {}
+    input: {
```
anakrish: You could tweak the yaml based test framework so that if `input`, and `want_result` are arrays, then the tester would use prepare_for_eval first and then eval the query using each input, want_result pair.

eric-therond: Good idea.

Another thing, previously an `eval` call was enough to evaluate all modules, now it's replaced by:
```
interpreter.prepare_for_eval(Some(&schedule), &None)?;
let results = interpreter.eval_modules(&input, true)?;
```
maybe we should still have an `eval` function doing the two previous steps?

Also what about having a boolean tracking the call of `prepare_for_eval`, this way if an user call  `eval_modules` (or `eval_module` or `eval_rule_with_input` or `eval_rule` or `eval_query_snippet`) without  `prepare_for_eval` before then we warn him (or we do a `prepare_for_eval` transparently for him?)

src/interpreter.rs
eric-therond: I will look at it.

tests/interpreter/cases/input/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+    x[a] {
+        a = input.x + 5
+    }
+"#;
+
+    let expected = Value::from_json_str(
+        r#" {
+            "test": {
+                "x": {
+                    "set!":[8, 15]
```
eric-therond: @anakrish result should only be "set!":[15] and not "set!":[8, 15]? We should clean the data on each evaluation?

[With OPA](https://pkg.go.dev/github.com/open-policy-agent/opa/rego#Rego.PrepareForEval) results are independent of each other on different evaluations.

tests/interpreter/mod.rs
eric-therond: I would say we need both? eval the first rule a second time (on an additional input), and then the query snippet also a second time?

tests/interpreter/cases/input/mod.rs
anakrish: If new input is supplied, we'd need to clean the data and internal evaluation state (if any). If the same input is used, we don't need to clean.

tests/interpreter/cases/input/tests.yaml
anakrish: > You could tweak the yaml based test framework so that if input, and want_result are arrays

It would be good to have a `multiple_inputs` flag as well. Only if that flag is specified as true, then `input` and `want_result` will be interpreted as arrays.

> eval function doing the two previous steps

That sounds good.

> or we do a prepare_for_eval transparently for him

We could start off with this approach and revisit it later if needed.

tests/interpreter/mod.rs
anakrish: Good point. We'd likely need to reevaluate all the rules and the query again, to be correct.

tests/interpreter/cases/input/tests.yaml
eric-therond: I was not able how to use a `multiple_inputs` flag to specify the type of `input` and `want_result` because a `Value` can be deserialized as an array, but maybe I am wrong and there is a way to do that?

Instead, I created an `Enum` and I modified each test case to specify what variant of the enum to expect.

eric-therond: I finally feel that warning the user about the mandatory `prepare_for_eval` call is better than calling `prepare_for_eval `transparently for the user because `prepare_for_eval `expects the data as argument but not `eval_ ` functions and so it changes a bit the logic if `eval_` takes data also as an argument.

src/interpreter.rs
eric-therond: Unfortunately I didn't succeed to use scopeguard; by doing something like that:
```
let _guard = scopeguard::guard(rule, |rule| {
      self.processed.insert(rule);
    });
```

I had a borrowing error.
I can try a bit more later.

eric-therond:
@anakrish I think the PR is ready to be reviewed
I've made changes to many files, but it's mostly to mark up yaml test cases to handle one or more inputs/results.

eric-therond:
@anakrish I don't know if you have time to review this PR ? can I do something more on it ?

anakrish:
Updated and checked in via https://github.com/microsoft/regorus/pull/26

Issue #36, opened 2023-11-05
simplify the regorus API


eric-therond:
Hello
I am facing troubles integrating regorus into an existing project mainly because of [references](https://github.com/microsoft/regorus/blob/main/src/interpreter.rs#L21) to `modules`/`schedule` in the `interpreter`.

Because of this and the `modules`/`schedule` lifetimes I need to refactor my project to call the regorus API (like in [test cases](https://github.com/microsoft/regorus/blob/main/tests/interpreter/mod.rs#L287)) and then to have a reference to the `interpreter` that is passed through functions of my project that need it (instead of storing the `interpreter` instance into a struct).

What do you think about "to hide" the complexity of the API (create the `Sources`, `Parse` and `Analyze` them, ...) and to store everything (struct instances not references) into the `interpreter`? something like that:
```
let mut interpreter = interpreter::Interpreter::new();
interpreter.add_module("myfile.rego");
interpreter.add_input("foo.json");
interpreter.eval_modules(enable_tracing)?;
```

anakrish:
Great idea. Do you want to take a stab at it?

eric-therond:
I tried a bit and the first part (storing `modules`, `schedule`, `source` objects in the struct interpreter...) requires a lot of changes in the whole code base and it's beyond my skills for the moment.

anakrish:
I've taken a pass at simplifying the API: https://github.com/microsoft/regorus/blob/d69b413c8efe38443ee297295f07f06e5b2e568c/examples/regorus.rs#L14.

Note, there are opportunities for optimizations and refactor/simplification/cleanup. The API will change in future.

anakrish:
@eric-therond 
Can you share your use-case? IIRC you explicitly call eval_rule.

eric-therond:
@anakrish 

1. At the beginning of the program, I parse and "compile" all the rego rules.
2. Then the program evaluates the rego rules (`eval_rule`) that are associated to group of target files (*.yaml => rego(rule1, rule2, rule3), *.json=> rego(rule4, rule5, rule6) ...).
3. In reality it's a bit more complex `eval_rule` can be trigerred depending on some "events".

I suppose I have to do use `eval_query`? I tried in the past, I don't remember why I was not able to continue in this direction. I will try again and update this thread.

anakrish:
Thank you. Do let me know how it turns out. I want to make the Engine API support your use case nicely as well.

PR #50, opened 2023-11-22
add full api to engine

eric-therond:
following #36
to move the rest of the user-API from interpreter to engine


src/engine.rs
```diff
@@ -6,73 +6,133 @@ use crate::interpreter::*;
...
-        Self::new()
-    }
+    interpreter: Interpreter,
+    prepared: bool,
 }
 
 impl Engine {
-    pub fn new() -> Self {
-        Self {
+    pub fn new() -> Result<Self> {
```
anakrish: In future we can change this so that all our `new` functions return T instead of Result<T>; `new` functions are expected to always succeed.

src/engine.rs
```diff
@@ -6,73 +6,133 @@ use crate::interpreter::*;
...
     pub fn clear_data(&mut self) {
-        self.data = Value::new_object();
+        self.interpreter.set_data(Value::new_object());
+        self.prepared = false;
     }
 
     pub fn add_data(&mut self, data: Value) -> Result<()> {
-        self.data.merge(data)
+        self.prepared = false;
+        self.interpreter.get_data().merge(data)
```
anakrish: `get_data_mut()` reads better.

src/engine.rs
```diff
@@ -6,73 +6,133 @@ use crate::interpreter::*;
...
     }
 
-    pub fn eval_query(&self, query: String, enable_tracing: bool) -> Result<QueryResults> {
-        // Analyze the modules and determine how statements must be scheduled.
-        let analyzer = Analyzer::new();
-        let schedule = analyzer.analyze(&self.modules)?;
+    pub fn prepare_for_eval(&mut self, enable_tracing: bool) -> Result<()> {
```
anakrish: Does this function need to be public?

src/interpreter.rs
```diff
@@ -86,16 +86,15 @@ struct LoopExpr {
 }
 
 impl Interpreter {
-    pub fn new(modules: &[Ref<Module>]) -> Result<Interpreter> {
+    pub fn new() -> Result<Interpreter> {
```
anakrish: In future we can change this so that Interpreter is returned instead of Result<Interpreter>

src/interpreter.rs
```diff
@@ -116,10 +115,30 @@ impl Interpreter {
...
+
+    pub fn set_functions(&mut self, functions: FunctionTable) {
+        self.functions = functions;
+    }
+
+    pub fn set_modules(&mut self, modules: &[Ref<Module>]) {
+        self.modules = modules.to_vec();
+    }
+
     pub fn get_modules(&mut self) -> &mut Vec<Ref<Module>> {
```
anakrish: `get_modules_mut` is a better name since a mutable reference is returned.

src/interpreter.rs
```diff
@@ -116,10 +115,30 @@ impl Interpreter {
...
         &mut self.modules
     }
 
+    pub fn set_init_data(&mut self, init_data: Value) {
+        self.init_data = init_data;
+    }
+
+    pub fn get_init_data(&mut self) -> &mut Value {
```
anakrish: `get_init_data_mut` is a better name since a mutable reference is returned.

Alternatively we could have a `merge_init_data(&mut self, v : Value)` that performs a merge of init data. That way we don't need to return mutable references.

anakrish (commented):
Overall, looks good to me.

src/interpreter.rs
eric-therond: Apparently I don't use `get_init_data` anylonger, I forgot to delete it.
There is a `set_init_data(&mut self, v : Value)` that does what you expect?

src/engine.rs
eric-therond: I think not, previously it was necessary for the end-user to explicitly calls this function and now it's done automatically behind the hood when calling `eval_*` functions of the engine. I don't see a valid reason for the end-user to call it.

anakrish (approved):
LGTM.

Note: Changing input would also require re-preparation.

eric-therond:
Can you explain why? I don't see anything in the `prepare_for_eval` function that depends on the input.

anakrish:
> Can you explain why? I don't see anything in the `prepare_for_eval` function that depends on the input.

I'm thinking a rule that depends on input values like below would need to be reevaluated when input changes
```
package test
a = input.a
```

Issue #72, opened 2023-12-24
[Lib] Implement missing `time` builtins


unexge:
Implement missing [time](https://www.openpolicyagent.org/docs/latest/policy-reference/#time) builtins with tests.

- [x] [time.add_date](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timeadd_date)
- [x] [time.clock](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timeclock)
- [x] [time.date](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timedate)
- [x] [time.diff](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timediff)
- [ ] [time.format](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timeformat)
- [x] [time.now_ns](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timenow_ns)
- [x] [time.parse_duration_ns](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timeparse_duration_ns)
- [ ] [time.parse_ns](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timeparse_ns)
- [x] [time.parse_rfc3339_ns](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timeparse_rfc3339_ns)
- [x] [time.weekday](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timeweekday)

Ensure that OPA tests (`cargo test --test opa -- time`) pass.

PR #82, opened 2023-12-31
Implement most of the builtin `time` module

unexge:
Partly addresses https://github.com/microsoft/regorus/issues/72.

With this PR we still fail to pass some OPA tests:
```
$ cargo test --test opa --features opa-testutil -- time # You need to pass `opa-testutil` for `test.sleep` function
...
OPA TESTSUITE STATUS
    FOLDER                                    PASS FAIL
    time                                    :   28    4

    TOTAL                                   :   28    4

MISSING FUNCTIONS
          FUNCTION                                 FAILURES
       1: time.parse_duration_ns                   1
          TOTAL                                    1
```
They fail because we can't parse Go's duration (`1m32s`) and time layout (`2006-01-02T15:04:05.999999999Z07:00`) at the moment.
I plan to address them in a follow-up PR.

### DIFF
--- a/src/builtins/time.rs
+++ b/src/builtins/time.rs
@@ -10,11 +10,13 @@ use crate::value::Value;
 use std::collections::HashMap;
 use std::time::SystemTime;
 
-use anyhow::{bail, Result};
-use chrono::{Datelike, Days, Months, NaiveDateTime};
+use anyhow::{anyhow, bail, Result};
+use chrono::{DateTime, Datelike, Days, FixedOffset, Local, Months, TimeZone, Timelike, Utc};
+use chrono_tz::Tz;
 
 pub fn register(m: &mut HashMap<&'static str, builtins::BuiltinFcn>) {
     m.insert("time.add_date", (add_date, 4));
+    m.insert("time.clock", (clock, 1));
     m.insert("time.now_ns", (now_ns, 0));
 }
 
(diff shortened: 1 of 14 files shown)


src/builtins/time.rs
```diff
@@ -3,28 +3,300 @@
...
+
+    Ok(Vec::from([
+        (datetime.year() as u64).into(),
+        (datetime.month() as u64).into(),
+        (datetime.day() as u64).into(),
+    ])
+    .into())
+}
+
+// Adapted from the official Go implementation:
```
anakrish: This function can be moved to a separate file that lists the original license.

anakrish (commented):
LGTM.

We'd need to include licenses as appropriate.

src/builtins/time.rs
unexge: I tried to address it with https://github.com/microsoft/regorus/pull/82/commits/56b3c7280a68ce57fbd92f932d6fbac6fdac1c34, but I am not sure if that's sufficient

anakrish: I believe that is good for now.

Later, we might have to declare a THIRD_PARTY_NOTICES file like https://github.com/openenclave/openenclave/blob/master/THIRD_PARTY_NOTICES
or find a way to avoid the code.

PR #105, opened 2024-01-12
Release preparation

unexge:
Updated `Cargo.toml` with the necessary fields, created [release-plz](https://github.com/MarcoIeni/release-plz) workflow file, and ran `release-plz update` to generate `CHANGELOG` file.

Release-plz GitHub Action quickstart guide (that also contains generating a token for crates.io) must be followed to complete the release automation: https://release-plz.ieni.dev/docs/github/quickstart.

Afterward, release-plz will create a rolling PR (example: https://github.com/MarcoIeni/release-plz/pull/1171) that will be updated with every change. Whenever we want to make a new release we will simply merge that PR and that will trigger GitHub workflow and run `release-plz release` which is going to detect the new version and then going to publish a release tag on GitHub and a new version on crates.io.

For the first release, we might need to run `release-plz release` manually.

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,0 +1,27 @@
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release-plz:
+    name: Release-plz
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
+        uses: MarcoIeni/release-plz-action@v0.5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,11 @@
 [package]
 name = "regorus"
+description = "A fast, lightweight Rego (OPA policy language) interpreter"
 version = "0.1.0"
 edition = "2021"
+license-file = "LICENSE"
+repository = "https://github.com/microsoft/regorus"
+keywords = ["interpreter", "opa", "policy-as-code", "rego", "confidential-computing"]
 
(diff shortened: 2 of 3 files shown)


.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
+        uses: MarcoIeni/release-plz-action@v0.5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
anakrish, line 27: Would this leak these secrets?

Cargo.toml
```diff
@@ -1,7 +1,11 @@
 [package]
 name = "regorus"
+description = "A fast, lightweight Rego (OPA policy language) interpreter"
 version = "0.1.0"
 edition = "2021"
+license-file = "LICENSE"
```
anakrish, line 6: Is there a way to include multiple licenses? Say of third-party code?

Cargo.toml
```diff
@@ -1,7 +1,11 @@
 [package]
 name = "regorus"
+description = "A fast, lightweight Rego (OPA policy language) interpreter"
```
anakrish, line 3: Does crates.io allow marking a release as alpha/beta?

anakrish (commented):
LGTM at a high level.

.github/workflows/release-plz.yml
unexge: GitHub redacts them: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#redacting-secrets-from-workflow-run-logs

Cargo.toml
unexge: Yes, we can add `-alpha` suffix, I'll change it to `0.0.1-alpha`

Cargo.toml
unexge: Cargo only supports one license file: https://github.com/rust-lang/cargo/issues/5933 but it does support multiple license expressions like: `license = "MIT OR Apache-2.0"`. I'll try to combine licenses in one license file.

anakrish: We want the project to be MIT licensed; only some borrowed parts to be differently licensed. We can leave the license file as-is till we find out if there is a precedent on how what we want is accomplished.

anakrish:
Merging this in to make progress.

anakrish (approved):
LGTM

PR #130, opened 2024-02-06
Partially implement Go's time format

unexge:
There are still some failing tests from Go's standard library tests (I was expecting it'd be much easier to parse Go's time format... 😄) but I believe this PR covers a good amount of them and also passes all OPA tests (`cargo test --test opa --features opa-testutil -- time`) so I think it's in a good state for a PR.

Fixes https://github.com/microsoft/regorus/issues/72

### DIFF
--- a/src/builtins/time.rs
+++ b/src/builtins/time.rs
@@ -12,8 +12,8 @@ use std::collections::HashMap;
 use anyhow::{anyhow, bail, Result};
 
 use chrono::{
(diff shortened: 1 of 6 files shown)


anakrish:
@unexge Thanks for the contribution! Yup, the `time` functions seem quite involved.

The following test-case fails. We can either support it or choose not to.

```rust
            parse_test_case(
                "custom: \"2006-01-02 15:04:05-07\"",
                "2006-01-02 15:04:05-07",
                "2010-02-04 21:00:57-08",
                true,
                false,
                1,
                0,
            ),
```

Issue #126, opened 2024-02-05
[binding] Java bindings


anakrish:
Implement JVM binding for Regorus

PR #147, opened 2024-02-17
Java bindings

unexge:
Part of https://github.com/microsoft/regorus/issues/126.

Java bindings for Regorus using [jni-rs](https://github.com/jni-rs/jni-rs). I adapted some build scripts from [OpenDAL's Java bindings](https://github.com/apache/opendal/tree/main/bindings/java).

I'll send follow-up PRs to set up publishing to Maven Central.


bindings/java/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+            System.out.println(resJson);
+        }
+    }
+}
+```
+
+and run it with:
+```shell
+$ java -cp target/regorus-java-0.0.1.jar:target/regorus-java-0.0.1-osx-aarch_64.jar Test.java
```
anakrish, line 47: For my own understanding, can java not figure out the `osx-aarch_64.jar` portion automatically?

anakrish (approved):
Great work! Thanks for your contribution!

bindings/java/README.md
unexge: I think not, it should be in the classpath somehow. I'm also not a Java expert and unsure if that's the optimal structure. Ideally, we should just have one fat JAR that contains everything but I'll need to research a bit and find the optimal structure. I'll work on it in the follow-up PRs.

PR #151, opened 2024-02-21
Java publishing

unexge:
Closes https://github.com/microsoft/regorus/issues/126.

Added GitHub workflow for publishing Java bindings to Maven Central. However, an account needs to be created by following: https://central.sonatype.org/register/central-portal/ and then secrets `OSSRH_USERNAME` and `OSSRH_TOKEN` needs to be populated for this workflow to be run (hopefully 😄 I haven't able to debug workflow script, but hoping it to run 🤞).

As per the discussion in https://github.com/microsoft/regorus/pull/147, this new workflow bundles native libraries for most popular OS, architecture pairs. So most customers just need to add a dependency to Regorus Java, and it will work without any extra effort. For other architectures, I added some instructions to README.


anakrish:
Thanks for the contribution, @unexge. Great work as usual!

PR #158, opened 2024-02-23
bindings/java: Link Linux libraries against glibc 2.17

unexge:
Since we're using a recent version of Ubuntu to build our libraries on Linux, they get linked against modern version of glibc. 
This is however causes problems on hosts with old glibc (I'm looking at you Amazon Linux 😄).

Luckily, we can use Zig and especially [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) to build our Linux libraries against old glibc.

Before this PR:

```bash
$ objdump -T x86_64-unknown-linux-gnu/libregorus_java.so | grep GLIBC | sed 's/.*GLIBC_\([.0-9]*\).*/\1/g' | sort -Vu
2.2.5
2.3
2.3.2
2.3.4
2.14
2.17
2.18
2.25
2.28
2.29
2.33
2.34
```

After this PR:

```bash
$ objdump -T x86_64-unknown-linux-gnu/libregorus_java.so | grep GLIBC | sed 's/.*GLIBC_\([.0-9]*\).*/\1/g' | sort -Vu
2.2.5
2.3
2.3.4
2.14
2.17
```

### DIFF
--- a/.github/workflows/publish-java.yml
+++ b/.github/workflows/publish-java.yml
@@ -15,9 +15,13 @@ jobs:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             extension: so
+            build_cmd: zigbuild
+            glibc: "2.17"
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-latest
             extension: so
+            build_cmd: zigbuild
+            glibc: "2.17"
           - target: x86_64-apple-darwin
             os: macos-latest
             extension: dylib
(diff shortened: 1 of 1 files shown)


anakrish:
Thanks for identifying and fixing this @unexge.

Would it be possible for you to share where/how you are using the Java binding. That will help us in making a case for publishing the bindings officially via the Microsoft pipelines.

unexge:
Hey @anakrish, I work on a security org at Amazon. We're using OPA/Rego to evaluate some security policies. Most of our services are currently in Java and have a pure Java Rego interpreter. We're looking for more performant alternatives. I made a PoC using Regorus with Java bindings internally and the performance was really promising, looking for the next steps now.

anakrish:
@unexge, Thanks for sharing your use case. It is a great data point for getting the bindings officially published.

Great to hear of your success with Regorus. I feel we can make it more performant. We've been mainly focused on compliance and understanding the nuances of Rego language so far.

eshepelyuk:
+1 for publishing jars to maven

Issue #172, opened 2024-03-08
Avoid deprecated warnings in use of chrono


anakrish:
Warnings like the following are now being generated:

```
warning: use of deprecated associated function `chrono::TimeDelta::seconds`: Use `TimeDelta::try_seconds` instead
   --> src/builtins/time/compat.rs:602:30
    |
602 |             ("5s", Duration::seconds(5)),
    |                              ^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated associated function `chrono::TimeDelta::seconds`: Use `TimeDelta::try_seconds` instead
   --> src/builtins/time/compat.rs:603:31
    |
603 |             ("30s", Duration::seconds(30)),
    |                               ^^^^^^^

```

@unexge Can you take a look to see how best to address the warnings?

PR #173, opened 2024-03-09
Don't use deprecated chrono `Duration` methods

unexge:
Some panicking methods on `Duration` are deprecated as part of chrono `0.4.35`. We switched to use `Duration::try_*` of those APIs.

Closes https://github.com/microsoft/regorus/issues/172


anakrish (approved):
LGTM

PR #186, opened 2024-03-25
`eval_rule`: Evaluate rules directly instead of queries

### LINKED ISSUE #185: Ability to evaluate rules directly instead of evaluating queries
OPA queries are similar to rule bodies or comprehension bodies. They can contain multiple rego expressions, multiple statements, and even loops. The [QueryResults](https://docs.rs/regorus/latest/regorus/struct.QueryResults.html) structure returned by evaluating the queries is also verbose - the value produced by each statement is captured along with the location information.

In many cases, such detail is not needed. All that is needed is the ability to execute and given rule and return the output.
Such an API would be faster than `eval_query'.

Here is another use case: https://github.com/microsoft/regorus/issues/36#issuecomment-1999177935 that favors evaluating rules directly instead of a query.

anakrish:
closes #185

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -208,6 +208,52 @@ impl Engine {
         &self.modules
     }
 
+    /// Evaluate rule(s) at given path.
+    ///
+    /// [`eval_rule`] is often faster than [`eval_query`] and should be preferred if
+    /// OPA style [`QueryResults`] are not needed.
+    ///
+    /// ```
+    /// # use regorus::*;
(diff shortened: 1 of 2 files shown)


anakrish:
@unexge Consider using `eval_rule` if possible. It is more performant than `eval_query` at the microsecond level.

unexge:
Thank you @anakrish! That improved our internal benchmarks between 10% to 30% 🚀

PR #188, opened 2024-03-29
Ruby bindings for existing FFI methods, plus eval_rule()

thedavemarshall:
I have a few things left to figure out, but I wanted to get some early feedback on my first time wring Rust. Feedback is welcome!

- [x] improve the Ruby tests
- [ ] add the Ruby bindings to the github workflows CI/CD

~~The other thing I'd call out is that I ran into some issues with a circular build when I had the `bindings/ruby/ext/regorusrb/Cargo.toml` depend on the relative location of `regorus`, so I locked it to `0.1.2` . Other language bindings seem to work with the relative paths, I think this might be something to do with the _other_ manifest at `bindings/ruby/Cargo.toml`? Not sure if resolving this is a blocker or if we can manually update the versions as needed for each release.~~
Update- I think I got it to work with relative paths for the crate!

The other callout is that rubygems expects a `Cargo.lock` file to be included, [source](https://github.com/rubygems/rubygems/blob/abb5d52f1e6a493758ef894dde16ec96274ca179/lib/rubygems/specification_policy.rb#L497-L504), which in this case is the top level regorus crate's Cargo.lock


thedavemarshall:

bindings/ruby/README.md
```diff
@@ -0,0 +1,101 @@
+# Regorusrb
+
+**Regorus** is
+
+  - *Rego*-*Rus(t)*  - A fast, light-weight [Rego](https://www.openpolicyagent.org/docs/latest/policy-language/)
+   interpreter written in Rust.
+  - *Rigorous* - A rigorous enforcer of well-defined Rego semantics.
+
+Regorus can be used in Ruby by  Use the bundler CLI to add the gem from remote git source:
```
anakrish: Seems like part of the sentence got removed somehow.

bindings/ruby/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+`rake build`
+
+then to install the gem and build the native extensions,
+
+`gem install --local ./pkg/regorusrb-0.1.0.gem`
+
+## Installation
+
+TODO: Replace `UPDATE_WITH_YOUR_GEM_NAME_IMMEDIATELY_AFTER_RELEASE_TO_RUBYGEMS_ORG` with your gem name right after releasing it to RubyGems.org. Please do not do it earlier due to security reasons. Alternatively, replace this section with instructions to install your gem from git if you don't plan to release to RubyGems.org.
```
anakrish: Not clear whom these instructions are intended for.

bindings/ruby/README.md
```diff
@@ -0,0 +1,101 @@
...
+    "/run/layers/p0-layer4",
+    "/run/layers/p0-layer5"
+  ],
+  "target": "/run/gcs/c/container0/rootfs"
+}
+
+engine.set_input(input)
+
+# Eval query
+results = engine.eval_query('data.framework.mount_overlay=x')
```
anakrish: Maybe we could show case eval_rule instead of eval_query?

bindings/ruby/ext/regorusrb/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+use regorus::Engine as RegorusEngine;
+use std::cell::RefCell;
+use std::cmp::Ordering;
+
+// `Value` exists under magnus, regorus, and serde_json, so be explicit
+
+#[derive(Default)]
+#[magnus::wrap(class = "Regorus::Engine")]
+pub struct Engine {
+    engine: RefCell<RegorusEngine>,
```
anakrish, line 11: Why do we need a RefCell?

bindings/ruby/ext/regorusrb/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    fn add_data_json(&self, json_string: String) -> Result<(), Error> {
+        let json_data: regorus::Value = serde_json::from_str(&json_string).map_err(|e| {
+            Error::new(runtime_error(), format!("Failed to parse JSON data: {}", e))
+        })?;
+
+        self.engine
+            .borrow_mut()
+            .add_data(json_data)
```
anakrish: Maybe we should use Engine::add_data_json?

bindings/ruby/ext/regorusrb/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    fn set_input_json(&self, json_string: String) -> Result<(), Error> {
+        let json_data: regorus::Value = serde_json::from_str(&json_string).map_err(|e| {
+            Error::new(
+                runtime_error(),
+                format!("Failed to parse JSON input file: {}", e),
+            )
+        })?;
+
+        self.engine.borrow_mut().set_input(json_data);
```
anakrish, line 122: Maybe we should use Engine::set_input_json?

anakrish (commented):
LGTM overall.
I don't quite follow the Ruby details; but will try it out locally.
The PR title could be updates as well.

anakrish:
Looks great, especially if this is the first time with Rust!

bindings/ruby/README.md
thedavemarshall: Ah, I've removed this text and revised the `Installation` section header.

I've reserved the `regorus` name in Rubygems by publishing an empty v0.1.0 gem there. Whenever we're ready, we can bump versions and I can transfer ownership in Rubygems to the appropriate account(s).

bindings/ruby/README.md
thedavemarshall: my mistake, updated now!

bindings/ruby/ext/regorusrb/src/lib.rs
thedavemarshall: Good catch, updated!

bindings/ruby/ext/regorusrb/src/lib.rs
thedavemarshall: Git missed here, but the `set_input_json` method now calls `engine.set_input_json` .

The `add_input_from_json_file` and `add_data_from_json_file` methods call the respective `engine.set_input` and `engine.add_data` methods, as the other language binding implementations appear to do.

bindings/ruby/ext/regorusrb/src/lib.rs
thedavemarshall: I think because we need an immutable struct to expose to Ruby, but we want a mutable reference to RegorusEngine to call add_data or eval_rule? I'm a Rust noob haha, doing my best to follow the magnus [docs](https://github.com/matsadler/magnus?tab=readme-ov-file#wrapping-rust-types-in-ruby-objects)

The other option of passing a `&mut self` as the first argument to those methods doesn't appear to work with the magnus macros.

anakrish: Makes sense. Thanks for the clarification.

bindings/ruby/test/test_regorus.rb
```diff
@@ -0,0 +1,180 @@
+# frozen_string_literal: true
```
anakrish, line 1: Nice tests.

anakrish (approved):
LGTM.

Thanks for this contribution!

anakrish:
I have created an issue #191.

thedavemarshall:
There are a few other things I haven't gotten to yet besides adding the Ruby bindings tests to the github actions, but I think these could each be followup PRs.

- [ ] custom Ruby exception class like `Regorus::Error` rather than using `RuntimeError` for more idiomatic error handling
- [ ] Support calling `eval_bool_query()` , `eval_allow_query()` , `eval_deny_query()` from Ruby
- [ ] Support calling `add_extension()` with Ruby code blocks - I have some ideas about this, but haven't had a chance to try anything out yet

@anakrish thank you so much for the feedback!

anakrish:
@thedavemarshall I will go ahead and merge the PR. The remaining items including the github action could be done as separate PRs.

anakrish:
> Support calling add_extension() with Ruby code blocks - I have some ideas about this

Interesting!

PR #201, opened 2024-04-10
Rewrite so that code compiles with chrono_tz 0.8.5 and 0.9.0

anakrish:
@unexge This is regarding #198 
I have rewritten the offending code so that it compiles with both 0.8.5 and 0.9.0.
Couple of questions:
1. Is there some idiomatic way to write this so that it compiles with both the versions?
2. Should we update chrono-tz or let it be? If we update, then we'd force a chrono-tz update on users of Regorus. If we don't, we might miss out on improvements that 0.9.0 might bring in. What would be your recommendation?

### DIFF
--- a/src/builtins/time.rs
+++ b/src/builtins/time.rs
@@ -9,7 +9,7 @@ use crate::value::Value;
 
 use std::collections::HashMap;
 
-use anyhow::{anyhow, bail, Result};
+use anyhow::{bail, Result};
 
 use chrono::{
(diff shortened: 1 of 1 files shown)


unexge:
I think your change would work for both 0.8.5 and 0.9.0. I think it's better to let it be in this version until they reach 1.0 and a stable API. We have to cope with breaking changes until then I think.

unexge (approved):

PR #225, opened 2024-05-05
Use alloc, core instead of std

anakrish:
- Replace std with alloc, core in most places in src Tests, bindings aren't changed.
- Introduce BuiltinsMap type alias inplace of HashMap. In no_std case, this could be aliases to BTreeMap
- Fix clippy warnings

### DIFF
--- a/src/builtins/sets.rs
+++ b/src/builtins/sets.rs
@@ -7,11 +7,11 @@ use crate::builtins::utils::{ensure_args_count, ensure_set};
 use crate::lexer::Span;
 use crate::value::Value;
 
-use std::collections::{BTreeSet, HashMap};
+use alloc::collections::BTreeSet;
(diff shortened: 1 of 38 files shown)


src/builtins/sets.rs
```diff
@@ -56,7 +56,7 @@ fn intersection_of_set_of_sets(
         };
 
         if first {
-            res = (**s).clone();
+            res.clone_from(&(**s));
```
mkulke: I think this can be just `res.clone_from(s)` it should auto-dereference.

mkulke (approved):
lgtm, small nit

src/builtins/sets.rs
anakrish: Great suggestion!

PR #231, opened 2024-05-09
add `std` feature

anakrish:
- `std` feature is enabled by default
- By default enable #![no_std] compilation
- Import std create if `std` feature is enabled or if testing
- Use core, alloc types
- Make it clear where std types are being used
- In no std, use BTreeMap in place of HashMap. HashMap is not available in no std due to lack of a secure random number generator

Note: The project does not yet compile without std feature being specified. But it's really close to being able to do so.


anakrish:
Merging to make progress.

PR #232, opened 2024-05-10
no_std support

anakrish:
- Disable default features in dependencies
- Use anyhow::Error::msg to map errors. Note: anyhow will itself be removed later.
- lazy_static/spin_no_std used in no_std environments
- ensure_no_std binary is built to target  thumbv7m-none-eabi to ensure that there are no std dependencies.  thumbv7m-none-eabi target has no std support.
- The opa-no-std feature enables only those Regorus features that work with no_std.


src/builtins/debugging.rs
```diff
@@ -43,7 +43,7 @@ pub fn print_to_string(
 // Additionally interpreter must allow undefined inputs to print.
 fn print(span: &Span, params: &[Ref<Expr>], args: &[Value], strict: bool) -> Result<Value> {
     let msg = print_to_string(span, params, args, strict)?;
-
+    let _ = core::convert::identity(&msg);
```
mkulke: why this noop? if this is to suppress a warning because of the feature flag:

```
    #[allow(unused_variables)]
    let msg = print_to_string(span, params, args, strict)?;
```

can be added (unused_mut for mutable variables).

However, since the whole "print" fn doesn't do anything in no-std, would it make sense to put the whole fn behind a std feature flag and dealing with it at the caller site?

src/parser.rs
```diff
@@ -74,6 +74,7 @@ impl<'source> Parser<'source> {
         let msg = format!(
             "`{kw}` will be treated as identifier due to missing `import future.keywords.{kw}`"
         );
+        let _ = core::convert::identity(&msg);
```
mkulke: similar remark as above about "print()" being a noop in no-std

src/builtins/debugging.rs
anakrish: Good suggestion. I refactored code to handle `print` differently. It is indeed a unique builtin in that undefined values are allowed, and we provide a `take prints` capability to make debugging msft/kata-containers easier.

src/parser.rs
anakrish: Done.

PR #239, opened 2024-05-18
Provide a way to obtain package names of loaded policies

### LINKED ISSUE #229: Provide a way to get the package name of a loaded rego file
Currently there is no direct way for the user to figure out package names defined in loaded policy files.

One way to obtain the package names is via a query:

```bash
$ regorus eval -d policy.rego "data[package_name]"
...
      "bindings": {
        "package_name": "agent_policy"
      }
...
```

```rust
let mut engine = Engine::new();
engine.add_policy(...);
engine.set_strict_builtin_errors(false); // Ensure that the query below does not raise errors due to missing data/input.

// Do not add_data since the added data shows up as keys in the `data` document.

// Create bindings to package names.
let results = engine.eval_query("data[package_name]".to_string(), false)?;
// Name of first package.
let package_name = results.result[0].bindings["package_name"].as_string()?.to_string();
```

The ask in this issue to provide a direct way.

See #228, #187

anakrish:
Engine::
- add_policy and add_policy_from_file return the package name.
- get_packages returns the names of loaded packages.

closes #229

### DIFF
--- a/examples/regorus.rs
+++ b/examples/regorus.rs
@@ -25,7 +25,7 @@ fn read_value_from_json_file(path: &String) -> Result<regorus::Value> {
     regorus::Value::from_json_str(&read_file(path)?)
 }
 
-fn add_policy_from_file(engine: &mut regorus::Engine, path: String) -> Result<()> {
+fn add_policy_from_file(engine: &mut regorus::Engine, path: String) -> Result<String> {
     #[cfg(feature = "std")]
     return engine.add_policy_from_file(path);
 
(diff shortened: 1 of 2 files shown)


anakrish:
CC @lquerel

@thedavemarshall You'd earlier asked for something like this. Does this help your use case?

lquerel:
Perfect. Thanks

thedavemarshall:
@anakrish this is awesome, thank you!

I've prepared some corresponding changes for the ruby bindings - https://github.com/microsoft/regorus/pull/240

PR #240, opened 2024-05-18
Update ruby bindings for add_policy to return package name

thedavemarshall:
Update ruby bindings for `#add_policy()` and `#add_policy_from_file` to return string package name.

Also bump the `rb_sys` runtime dependency, and the development dependencies.


thedavemarshall:
This PR depends on the changes in https://github.com/microsoft/regorus/pull/239

anakrish (approved):
Thanks for proactively keeping the binding up to date!

Issue #242, opened 2024-05-18
Build and test bindings in CI


anakrish:
Currently, the bindings are not regularly built in CI. This can lead to the bindings going out of date. We need to build the bindings with each PR job.

Ideally, there needs to be a way to test the bindings too. But not sure if that will bring in a lot of dependencies.

Add github workflows to test

- [x] C/C++ binding
- [x] C no_std binding
- [x] C# binding
- [x] Go binding
- [x] Java binding
- [x] Python bindings
- [x] Ruby bindings
- [x] WASM bindings

PR #244, opened 2024-05-18
Add test-ruby CI for github actions

thedavemarshall:
This change adds a `test-ruby` CI workflow that is triggered on pushes and pull requests to `main`, and will run the default rake task of `compile test rubocop`.

Happy to change the naming or triggers here, but wanted to see what it would take to get the ruby test suite run as part of CI.

see https://github.com/oxidize-rb/actions/blob/main/setup-ruby-and-rust/readme.md for docs on the rust and ruby setup action.

see https://github.com/thedavemarshall/regorus/actions/runs/9142734035/job/25138614189 for an example run of this workflow.

### DIFF
--- a/.github/workflows/test-ruby.yml
+++ b/.github/workflows/test-ruby.yml
@@ -1,0 +1,30 @@
+name: test-ruby
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Ruby and Rust
+        uses: oxidize-rb/actions/setup-ruby-and-rust@v1
+        with:
+          ruby-version: "3.3.1"
+          rustup-toolchain: "stable"
+          bundler-cache: true
+          cargo-cache: true
+          working-directory: "bindings/ruby"
+
+      - name: Run ruby tests
+        run:  |
+          cd bindings/ruby
+          bundle exec rake


thedavemarshall:
@anakrish Is this the approach you were thinking for https://github.com/microsoft/regorus/issues/242 ?

anakrish:
@thedavemarshall Looks great! Thanks for taking this on.

One question: How trustworthy is `oxidize-rb/actions/setup-ruby-and-rust@v1`?

thedavemarshall:
@anakrish Great question! I'm not a security expert or rust expert, and I'm also relatively new to GitHub actions, but here's what I've found. The summary is that _I_ would trust this action, but I'm happy to try a different approach if you'd prefer!

- the [oxidize-rb/actions](https://github.com/oxidize-rb/actions) repo with the `oxidize-rb/actions/setup-ruby-and-rust@v1` source code is maintained by the same individuals in the [oxidize-rb GitHub organization](https://github.com/oxidize-rb) as https://github.com/oxidize-rb/rb-sys , which is used as a runtime dependency for the Ruby-rust bindings of Regorus (https://rubygems.org/gems/rb_sys and https://crates.io/crates/rb-sys) , and appears to be the recommended way to CI test ruby-rust bindings https://github.com/oxidize-rb/rb-sys?tab=readme-ov-file#quick-links
- Several other ruby gems with rust native extensions use this action, including
  - https://github.com/gjtorikian/commonmarker/blob/main/.github/workflows/test.yml#L22
  - https://github.com/gbmoretti/ratatui_rb/blob/main/.github/workflows/main.yml#L22
  - https://github.com/taylorthurlow/rust_json_schema-rb/blob/main/.github/workflows/main.yml#L22
  - https://github.com/yammine/roaring-rb/blob/main/.github/workflows/ci.yml#L27
  - https://github.com/adampetro/bluejay-rb/blob/main/.github/workflows/ci.yml#L24
  - https://github.com/jinshuju/rahocorasick/blob/main/.github/workflows/release-gem.yml#L22
  - https://github.com/bytecodealliance/wasmtime-rb/blob/main/.github/workflows/ci.yml#L51
  - https://github.com/wilsonzlin/minify-html/blob/master/.github/workflows/ruby.yml#L46
  - https://github.com/huacnlee/rucaptcha/blob/main/.github/workflows/build.yml#L12
- the [repo](https://github.com/oxidize-rb/actions/blob/main/setup-ruby-and-rust/action.yml) with the actions is MIT licensed, so we could just copy the source we need and define a minimal Ruby-rust CI action in this repo alongside the Regorus source code?
- https://github.com/huacnlee/autocorrect/tree/main Here's one example I found of a rust library with ruby bindings that has CI _without_ using that action, https://github.com/huacnlee/autocorrect/actions/runs/9153054739/job/25161407807 , so I could try to hand roll something similar if you'd prefer that approach

anakrish:
@thedavemarshall Thanks for the investigation. Based on your findings, I agree with you that it is reasonable to trust `oxidize-rb` since we already trust `rb-sys`. Additionally, `magnus` also uses `oxidize-rb` in its CI:

https://github.com/matsadler/magnus/blob/324f57e44da5948ea146639d0c2330bc26c8b911/.github/workflows/test.yaml#L50

Also, `bytecodealliance` in your list is focused on security.

Based on all the above, I'm OK to trust `oxidize-rb`.

anakrish (approved):
LGTM

anakrish:
@thedavemarshall This PR inspired me to do #247.

Issue #237, opened 2024-05-17
publish-java action has never been run, bindings are not in Maven


bromls:
The documentation in `README.md` and in `bindings/java/README.md` indicates that Java bindings for `regorus` are published to Maven Central. I don’t see the bindings in Maven, and the GitHub action for making this true appears to have never been run: https://github.com/microsoft/regorus/actions/workflows/publish-java.yml

anakrish:
@bromls This is a documentation issue. CC @unexge, the owner of the Java binding.

As noted in the [bindings](https://github.com/microsoft/regorus?tab=readme-ov-file#bindings), we currently don't publish bindings to repositories due to the operational overhead.
However, I have been collecting use cases to help justify the cost of publishing the bindings. Would it be possible to share your use case?

anakrish:
Addressed by #246

PR #246, opened 2024-05-20
Update `README.md` for Java bindings to mention we don't publish to Maven Central currently

unexge:
See https://github.com/microsoft/regorus/issues/237


anakrish (approved):
LGTM

PR #247, opened 2024-05-21
Fix bindings and add CI tests

### LINKED ISSUE #242: Build and test bindings in CI

anakrish:
closes #242 

Also does some related cleanup. The PR workflow is split into separate workflows.

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -1,0 +1,37 @@
+name: bindings/python
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+env:
+  PYTHON_VERSION: "3.10"
+    
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          architecture: x64
+
+      - name: Build wheels
+        uses: PyO3/maturin-action@63b75c597b83e247fbf4fb7719801cc4220ae9f3 # v1.43.0
+        with:
+          target: x86_64
+          args: --release --out dist --manifest-path bindings/python/Cargo.toml
+          sccache: 'true'
+
+      - name: Test wheel
+        run:  |
+          pip3 install dist/regorus-*.whl
+          cd bindings/python
+          python3 test.py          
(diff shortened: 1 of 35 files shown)


.github/workflows/test-python.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          architecture: x64
+
+      - name: Build wheels
+        uses: PyO3/maturin-action@v1
```
mkulke:
```suggestion
        uses: PyO3/maturin-action@6c20e5e8673d2470a20e14adf067b68db6fee644 # v1.43.0
```

Pinning to hashes for workflows by non-verified authors is [recommended](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

mkulke (commented):
small nit

.github/workflows/test-python.yml
anakrish: Great suggestion!

bindings/ruby/ext/regorusrb/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "regorusrb"
-version = "0.1.0"
```
thedavemarshall, line 3: I don't think it matter at this point since the ruby bindings aren't being published anywhere, but the Ruby gem also has a version specified in https://github.com/microsoft/regorus/blob/main/bindings/ruby/lib/regorus/version.rb#L4 that gets used in the gemspec.

thedavemarshall (approved):
@anakrish This is awesome!

bindings/ruby/ext/regorusrb/Cargo.toml
anakrish: Thanks for point that out. I will address it in the next PR to keep the bindings up to date.

anakrish: @thedavemarshall Can you submit a PR to update the version number in `version.rb`. I tried to do it, but that caused an error due to something being locked.

Issue #254, opened 2024-05-29
Provide a way to obtain policy file names and content from an Engine


bromls:
To support use cases like policy auditing, it would be helpful if the `Engine` exposed a function that allowed callers to iterate over the mappings of policy file names to policy file content.
An application that uses Rego policies may be run with different policies over time, as the software or configuration is updated, or based on the runtime environment.
For use cases like “[Collaboration using Import](https://www.openpolicyagent.org/docs/latest/faq/#collaboration-using-import)”, the policy Engine supports composition of packages, and accepts multiple policy files.
All of this information is available during Engine creation, as it is input to the `add_policy` and `add_policy_from_file` functions. However, some applications may separate `Engine` creation from where an `Engine` is used to evaluate rules or queries. Therefore, it would be valuable for an `Engine` user to have access to the policy file content.

anakrish:
@bromls How do policies get audited? Do you compute the hash of the policy file contents?

PR #255, opened 2024-05-30
expose policy info from the engine

thedavemarshall:
This change is to expose more package data from the engine, potentially addressing https://github.com/microsoft/regorus/issues/254

@anakrish I know you've mentioned the ast and some of the internals are unstable and being iterated on, so I'd love you input here if this is a good approach, or if you're not ready to expose these things through the engine.

I had some trouble getting a method of `package name -> [rule_name1, rule_name2]` , so I'm curious if there's a good way to do that with the current abstractions, or if that's something I could potentially help out with.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -129,6 +129,61 @@ impl Engine {
             .collect()
     }
 
+    pub fn get_packages_texts(&self) -> Result<Map<String, String>> {
+        let mut packages_text: Map<String, String> = Map::new();
+
+        for module in &self.modules {
+            let path_string = Interpreter::get_path_string(&module.package.refr, Some("data"))?;
+            let package_text = module.package.span.text().to_string();
+            packages_text.insert(path_string, package_text);
+        }
+
+        Ok(packages_text)
+    }
+
+    pub fn get_packages_imports(&self) -> Result<Map<String, Vec<String>>> {
+        let mut packages_imports: Map<String, Vec<String>> = Map::new();
+
+        for module in &self.modules {
+            let path_string = Interpreter::get_path_string(&module.package.refr, Some("data"))?;
+            let import_strings: Result<Vec<String>> = module
+                .imports
+                .iter()
+                .map(|import| Ok(import.span.text().to_string()))
+                .collect();
+            packages_imports.insert(path_string, import_strings?);
+        }
+
+        Ok(packages_imports)
+    }
+
+    pub fn get_packages_policies(&self) -> Result<Map<String, Vec<String>>> {
+        let mut packages_policies: Map<String, Vec<String>> = Map::new();
+
+        for module in &self.modules {
+            let path_string = Interpreter::get_path_string(&module.package.refr, Some("data"))?;
+            let policy_strings: Result<Vec<String>> = module
+                .policy
+                .iter()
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -129,6 +129,61 @@ impl Engine {
             .collect()
     }
 
+    pub fn get_packages_texts(&self) -> Result<Map<String, String>> {
```
anakrish, line 132: Given that these APIs are public, they would need to be documented as well as tested via doc tests.

anakrish, line 132: For my own understanding, what is the use of such an that maps filenames to package names?

src/engine.rs
```diff
@@ -129,6 +129,61 @@ impl Engine {
...
+        for module in &self.modules {
+            let path_string = Interpreter::get_path_string(&module.package.refr, Some("data"))?;
+            let package_text = module.package.span.text().to_string();
+            packages_text.insert(path_string, package_text);
+        }
+
+        Ok(packages_text)
+    }
+
+    pub fn get_packages_imports(&self) -> Result<Map<String, Vec<String>>> {
```
anakrish, line 144: Note that there is also a `as` part: 
```rego
import data.x as y
```

anakrish (commented):
@thedavemarshall Thank you for taking this on!

I understand #254 to want the Engine to provide a way to obtain entire policy file contents. I don't think the PR addresses that yet. 

> @anakrish I know you've mentioned the ast and some of the internals are unstable and being iterated on, so I'd love you input here if this is a good approach, or if you're not ready to expose these things through the engine.

I have been thinking about exposing the AST. In addition to the AST likely being changed in some time, the other challenge is how we can expose the AST to the bindings. The AST nodes are going to be cumbersome to expose to each language.

Some thoughts:
- We could expose AST publicly via a cargo feature. That will allow it to be used and refined based on feedback.
- We could version the AST.
- Instead of ad-hoc functions should we just define a mapping of AST nodes to json? We could then serialize entire AST to json and make it available to the different bindings as well.

I'd love your feedback since you have a concrete usecase.

thedavemarshall:
@anakrish thanks for the thoughtful feedback!

> I understand #254 to want the Engine to provide a way to obtain entire policy file contents. I don't think the PR addresses that yet.

ah I think I misunderstood when I first read those requirements, my mistake!

>     * We could expose AST publicly via a cargo feature. That will allow it to be used and refined based on feedback.
yeah, I think AST exposure is a great candidate for a rust crate feature

>     * We could version the AST.
I like this idea! Or at least versioning the JSON output?

>     * Instead of ad-hoc functions should we just define a mapping of AST nodes to json? We could then serialize entire AST to json and make it available to the different bindings as well.
I think this is similar to how opa does it for their policy HTTP API, [docs](https://www.openpolicyagent.org/docs/latest/rest-api/#policy-api)

> I'd love your feedback since you have a concrete usecase.
I think exposing a JSON representation of the AST is a good solution. Tomorrow I'll take an in-depth look at https://github.com/microsoft/regorus/pull/266 , which at first glance seems a great fit for my use case!

Closing this PR in favor of https://github.com/microsoft/regorus/pull/266

PR #258, opened 2024-05-30
Revert "chore: release v0.2.0"

anakrish:
Reverts microsoft/regorus#257


jdgo-mars:
Hi, first of all thank you for the work you do!

jdgo-mars:
Hi, first of all thank you for the work you do
since there is no context here

jdgo-mars:
Hi, first for all thank you for the work you do!
Regarding `0.2.0` release: after this revert, bindings are not in sync with current version.
Is there a specific reason for this?
Thank you!

anakrish:
@jdgo-mars, Thanks for reporting it. There is no special reason for it. At the time, that was the quickest way for me to get to a stage to publish v0.2.0.

I will manually fix the version numbers of extensions for now.
Going forward we need some way to update the version numbers consistently even though we only publish the main regorus crate.

PR #266, opened 2024-06-06
Provide ability to get JSON representation of policy AST

### LINKED ISSUE #265: Allow ability to introspect AST of policies
The goal of this issue is to provide a read-only view of the AST of loaded policies. The could for example be accomplished by serializing the AST to JSON. This approach allows keeping the AST data structure private and amenable to change while still retaining the JSON mapping.

CC: @thedavemarshall

anakrish:
closes #265

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ doctest = false
 default = ["full-opa", "arc"]
 
 arc = ["scientific/arc"]
+ast = []
 base64 = ["dep:data-encoding"]
 base64url = ["dep:data-encoding"]
 coverage = []
--- a/bindings/ffi/Cargo.toml
+++ b/bindings/ffi/Cargo.toml
@@ -13,7 +13,8 @@ regorus = { path  = "../..", default-features = false }
 serde_json = "1.0.113"
 
 [features]
-default = ["std", "coverage", "regorus/arc", "regorus/full-opa"]
+default = ["ast", "std", "coverage", "regorus/arc", "regorus/full-opa"]
+ast = ["regorus/ast"]
 std = ["regorus/std"]
 coverage = ["regorus/coverage"]
 custom_allocator = []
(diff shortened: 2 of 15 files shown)


anakrish:
@thedavemarshall Can you try out this PR to see if it helps your usecase?

thedavemarshall:
@anakrish this is super useful, thank you!

This unblocks the use case of deconstructing REGO documents into a format that can be easily be manipulated by a Ruby application!

Is there a recommended way load a JSON AST representation directly into the engine? Else I can make a simple parser in Ruby to turn the AST JSON representation into a REGO document and then load that into the engine. Either way, this is great!


I sketched out a couple tests for the ruby biniding `get_ast_as_json_ast` . If you think the AST representation is stable enough to assert against, I can open a PR to add https://github.com/microsoft/regorus/commit/595cb2eed0e7818aaeda55ce5573248eda60b292 , or feel free to copy/paste those tests into this PR!

anakrish:
@thedavemarshall Thanks for trying out the PR. Great that it helps with the Ruby use case.

I have also added a `version` field. I will go ahead and merge it to make progress. We can refine it in subsequent PRs.

> Is there a recommended way load a JSON AST representation directly into the engine?

Currently no. Many AST nodes store a `Span` and the span holds a reference to the `Source`.  To avoid duplication, we don't serialize the `Source` with each span. To handle this, we need a special deserializer than can set the missing `Source` reference correctly.

Also, I'm not sure if the `Span` can be created accurately unless the AST is created via parsing.

> Else I can make a simple parser in Ruby to turn the AST JSON representation into a REGO document and then load that into the engine.

Generating Rego from the AST and then supplying it back to the engine is a good approach. 

> I sketched out a couple tests for the ruby biniding get_ast_as_json_ast . If you think the AST representation is stable enough to assert against, I can open a PR to add https://github.com/microsoft/regorus/commit/595cb2eed0e7818aaeda55ce5573248eda60b292

Yes, please submit a PR.

PR #268, opened 2024-06-18
rename method from Clone to clone inEngine class to match the java naming convention and definition in the java.lang.Object.

balcanuc:
rename method from `Clone` to `clone` in `Engine` class to match the java naming convention and definition in the `java.lang.Object`.

An alternative would be to mark the current `Clone` as `@Deprecated` and add a new method `clone`. As the fails happens on compile time and Regorus is not at a peak adoption,  I think that just renaming the method is better.


anakrish:
@balcanuc Thanks for the contribution. Can you rebase to the latest main to unblock testing this PR?

anakrish (approved):
LGTM.

anakrish:
Merging this since the blocking clippy issue already fixed in main and the rebased pr passes locally.

PR #277, opened 2024-07-11
adding support for net4.0 in .NET and extension list as an example

yangjie-msft:
created an example of agent extension policy that is used for Azure agent extensions. 
added support for .NET framework 4.0 binding. 
Reorganized the structure accordingly due to the above changes.


anakrish:
@yangjie-msft Thanks for the contribution. Kindly squash you commits to a single changelist.

PR #279, opened 2024-07-12
add extension_list example

yangjie-msft:
created an example of agent extension policy that is used for Azure agent extensions.
added support for .NET framework 4.0 binding.
added a pytest test for the policy using the python bindings (thus also tested python binding)
Reorganized the structure accordingly due to the above changes.


bindings/csharp/net40/Program.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
```
mgunnala, line 1: should we add a proper C# test suite before merging to main?

yangjie-msft: we could do it in later stages. there are a couple of things we can do but let's first have the base line into the main.

mgunnala (approved):

.github/workflows/test-csharp.yml
```diff
@@ -19,8 +19,8 @@ jobs:
 
       - name: Build
         run: dotnet build
-        working-directory: ./bindings/csharp
+        working-directory: ./bindings/csharp/net8.0
```
anakrish, line 22: We could test both net8.0 and net40

bindings/csharp/net40/Program.cs
anakrish, line 1: Add copyright notice in each file

examples/extension_list/agent_extension_policy.rego
```diff
@@ -0,0 +1,122 @@
+package agent_extension_policy
```
anakrish, line 1: Add copyright notice for each source file.

anakrish (commented):
Looks good for most part.

It looks like net40 is not being used. We should either lock it down with tests of remove it altogether if it is not needed.

yangjie-msft:
> Looks good for most part.
> 
> It looks like net40 is not being used. We should either lock it down with tests of remove it altogether if it is not needed.

net40 is not tested because I am not sure if the tool chain can support net8.0 and net40 at the same time as net40 is pretty old but still needed by some applications.

.github/workflows/test-csharp.yml
yangjie-msft: I can try, but not sure if the tool chain can support both of them at the same time

examples/extension_list/agent_extension_policy.rego
yangjie-msft: will do.

bindings/csharp/net40/Program.cs
yangjie-msft: will do.

PR #281, opened 2024-07-15
add extension_list example

yangjie-msft:
- Created an example of extension policy
- Added C# binding support of .NET framework 4.0 and created a Nuget spec for it.
- Added a pytest in python bindings to test the extension policy and the python binding
- Restructured the example and Csharp binding directories due to above changes.
- Added copyrights.
- Added a Windows workflow for .NET 4.0 build and test.


anakrish (approved):
LGTM

Thanks for locking down the trusted extensions usecase.

PR #283, opened 2024-07-20
build.rs create hooks dir if not exists

sneivandt:
build.rs should create the `.git/hooks` directory if it does not exist, to address the following error when attempting to copy the hooks.

```
error: failed to run custom build command for `regorus v0.2.1 (C:\Source\regorus)`

Caused by:
  process didn't exit successfully: `C:\Source\regorus\target\debug\build\regorus-1728766aaa50ee3c\build-script-build` (exit code: 1)
  --- stderr
  Error: The system cannot find the path specified. (os error 3)
```


anakrish (approved):
LGTM

Thanks for the contribution @sneivandt

PR #295, opened 2024-08-08
fix: Match OPA behavior for split

### LINKED ISSUE #291: Strings::split incompatibility for empty delimiters
OPA and Regorus evaluate split differently when empty delimiters are
used.

In particular, the OPA implementation seems to follow the Go
strings.Split definition ("If sep is empty, Split splits after each
UTF-8 sequence"). In comparison, Regorus uses str::split, which also
returns the beginning and end of the string.

For example:
```
$ opa eval 'split("test", "")'
{
  "result": [
    {
      "expressions": [
        {
          "value": [
            "t",
            "e",
            "s",
            "t"
          ],
          "text": "split(\"test\", \"\")",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

vs.

```
$ regorus eval 'split("test", "")'
{
  "result": [
    {
      "expressions": [
        {
          "value": [
            "",
            "t",
            "e",
            "s",
            "t",
            ""
          ],
          "text": "split(\"test\", \"\")",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

anakrish:
In case of empty delimiter, Rust's split returns leading and trailing empty strings whereas Golang's doesn't.
Change behavior to match Golang/OPA.

fixes #291

### DIFF
--- a/src/builtins/strings.rs
+++ b/src/builtins/strings.rs
@@ -146,11 +146,18 @@ fn split(span: &Span, params: &[Ref<Expr>], args: &[Value], _strict: bool) -> Re
     let s = ensure_string(name, &params[0], &args[0])?;
     let delimiter = ensure_string(name, &params[1], &args[1])?;
 
-    Ok(Value::from_array(
+    // Handle https://github.com/microsoft/regorus/issues/291
+    let parts: Vec<Value> = if delimiter.as_ref() == "" {
+        // If delimiter is "", str::split returns a leading and trailing "" whereas Golang's split doesn't.
+        // Therefore avoid str::split and instead return each char as a Value::String.
+        s.chars().map(|c| Value::from(c.to_string())).collect()
+    } else {
         s.split(delimiter.as_ref())
             .map(|s| Value::String(s.into()))
-            .collect(),
-    ))
+            .collect()
+    };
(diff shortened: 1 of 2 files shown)


src/builtins/strings.rs
```diff
@@ -146,11 +146,32 @@ fn split(span: &Span, params: &[Ref<Expr>], args: &[Value], _strict: bool) -> Re
...
-    ))
+    let mut parts: Vec<Value> = s
+        .split(delimiter.as_ref())
+        .map(|s| Value::String(s.into()))
+        .collect();
+
+    // Handle https://github.com/microsoft/regorus/issues/291
+    // If delimiter is "", rust returns a leading and trailing "", Golang doesn't.
+    if delimiter.as_ref() == "" {
+        if parts.len() >= 2 {
```
codido: Any chance this could be simplified, and just set parts to `s.chars().collect()` in this case?

codido (approved):

src/builtins/strings.rs
anakrish: Great suggestion. Fixed.

PR #355, opened 2024-12-31
resolve anyhow compile errors

thedavemarshall:
@anakrish while attempting to run the tests from `main` on a fresh development environment, I was getting several compiler warnings and errors. This intends to be the smallest change to get back to green.

- add missing `--features` flag for `aci rego-extensions` and `kata rego-extensions` commands in `scripts/pre-push`
- I'm not entirely clear on the traits for why `?` wasn't playing well with `anyhow`, maybe it has something to do with `default-features = false` and `no_std` flag option? the fix seems to be `map_err()` calls
- replace unnecessary lifetime references `impl<'a> GoTimeFormatItems<'a>` with `impl GoTimeFormatItems<'_>`, similar for `impl Clone for Box<dyn '_ + Extension> {`

These changes allow me to compile the project and run tests without error or warning output.

### DIFF
--- a/scripts/pre-push
+++ b/scripts/pre-push
@@ -32,8 +32,8 @@ if [ -f Cargo.toml ]; then
 
    # Ensure that all tests pass with extensions
    cargo test -r --features rego-extensions
-   cargo test -r --test aci rego-extensions
-   cargo test -r --test kata rego-extensions
+   cargo test -r --test aci --features rego-extensions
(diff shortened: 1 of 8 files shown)


anakrish:
@thedavemarshall Thanks for looking into this. This unblocks couple of PRs submited by depandabot.

anakrish (approved):
LGTM

PR #383, opened 2025-03-18
feat: Regorus nuget package

anakrish:
Organize C# binding example into separate Regorus nuget package and a test app.

The nuget package targets netstandard 2.0 and 2.1. The test app is tested for netframework 8.0.

Also remove net40 example. Can be added back later if needed.


bindings/csharp/Regorus/Regorus.cs
```diff
@@ -0,0 +1,305 @@
+using System.Runtime.InteropServices;
+using System.Text;
+using System;
+
+#nullable enable
+namespace Regorus
+{
+	public sealed class Exception : System.Exception
```
anthony-c-martin: I would either rename this (e.g. `RegorusException` or something more specific), or just remove it and instead throw the built-in `System.Exception`. Otherwise when consumers import the `Regorus` namespace, they may run into naming conflicts if they reference the existing `Exception` class.

bindings/csharp/Regorus/Regorus.cs
```diff
@@ -0,0 +1,305 @@
+using System.Runtime.InteropServices;
```
anthony-c-martin, line 5: [nit] standard convention for C# is 4 spaces rather than tabs for whitespace

bindings/csharp/Regorus/Regorus.cs
```diff
@@ -0,0 +1,305 @@
...
+
+#nullable enable
+namespace Regorus
+{
+	public sealed class Exception : System.Exception
+	{
+		public Exception(string? message) : base(message) { }
+	}
+
+	public sealed class Engine : System.IDisposable
```
anthony-c-martin: Not sure if there's any downside to just marking the whole class as `unsafe`, to avoid having to wrap individual blocks in `unsafe`.

```csharp
public sealed unsafe class Engine : IDisposable
{
```

bindings/csharp/Regorus/Regorus.cs
```diff
@@ -0,0 +1,305 @@
...
+		~Engine() => Dispose(disposing: false);
+
+		public Engine Clone()
+		{
+			var clone = (Engine)this.MemberwiseClone();
+			unsafe
+			{
+				clone.E = Regorus.Internal.API.regorus_engine_clone(E);
+			}
+			return clone;
```
anthony-c-martin: I've not seen `MemberwiseClone` used very often. IMO a more idiomatic option to support cloning would be to define a private constructor overload:

```csharp
private unsafe Engine(Internal.RegorusEngine* engine)
{
    this.E = engine;
}

public unsafe Engine()
    : this(Regorus.Internal.API.regorus_engine_new())
{
}

public unsafe Engine Clone()
    => new(Internal.API.regorus_engine_clone(E));
```

.github/workflows/test-csharp.yml
```diff
@@ -17,10 +17,27 @@ jobs:
...
+      - name: Pack Regorus nuget
+        run: dotnet pack; unzip -l bin/Release/Regorus*.nupkg
+        working-directory: ./bindings/csharp/Regorus
+        
+      - name: Restore Test App
+        run: dotnet restore /p:RestoreSources=./../Regorus/bin/Release
+        working-directory: ./bindings/csharp/TestApp
+
+      - name: Build Test App 8.0
+        run: dotnet build --framework net8.0
```
anthony-c-martin: Might be worth running tests against different platforms - it's very straightforward to do this with `matrix` in GitHub workflows. Here's an example: https://github.com/Azure/bicep/blob/c5d378c4ba2e73a79d6ae105f41997fa61003d3a/.github/workflows/build.yml#L569-L578

bindings/csharp/Regorus/Regorus.csproj
```diff
@@ -0,0 +1,62 @@
...
+
+
+  <!-- Determine build os/arch -->
+  <PropertyGroup>
+    <IsLinuxX64>false</IsLinuxX64>
+    <IsWindowsArm64>false</IsWindowsArm64>
+    <IsWindowsX64>false</IsWindowsX64>
+    <IsOSXArm64>false</IsOSXArm64>
+
+    <IsLinuxX64 Condition="$([MSBuild]::IsOsPlatform('Linux')) And $([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture) == X64">true</IsLinuxX64>
```
anthony-c-martin: Isn't this going to vary the config depending on which platform/architecture that the **build machine** is running?

I'd have thought you'd want consitent results wherever the build runs to get either:
* Single nuget package with multiple architecture libaries bundled inside and a means of resolving the correct one
* A nuget package per platform architecture

bindings/csharp/Regorus/Regorus.cs
anakrish: Fixed. Needed to tweak a few vscode settings.

bindings/csharp/Regorus/Regorus.cs
anakrish: Good point. Removed this Exception class and used System.Exception instead.

bindings/csharp/Regorus/Regorus.cs
anakrish: Nice idea! That simplified the code a whole lot.

bindings/csharp/Regorus/Regorus.cs
anakrish: Nice suggestion.

.github/workflows/test-csharp.yml
anakrish: Thanks for the nudge. Done.

bindings/csharp/Regorus/Regorus.csproj
anakrish: The original intent was to support both a pipleine build for all supported targets and also a local build for only the current host platform for local development workflows. The host-target-triple was being figured out was so that `cargo build -r --target <host-target-triple>` can be invoked to generate the regorus ffi shared library in `target/<host-target-triple>/release`, just as the pack expects. If host-target-triple was not passed to cargo, then the shared library will be generated in `target/release` and won't get picked up by pack.

I have removed the local workflow completely to simpify the csproj.

PR #391, opened 2025-04-09
Initial draft of Regorus extension to invoke C# code

kusha:
It is still early draft that shows invoke of C# sharp code within extension during Rego policy evaluation. Example policy is in `bindings/csharp/scripts/invoke.csx`

TODO:
- [ ] Implement it as a built-in extension, instead of placing it in FFI
- [ ] Decide if not a thread-safe solution to store callbacks mapping is sufficient
- [ ] Validate typing support to callback payload and result

### DIFF
--- a/bindings/csharp/Regorus/Regorus.cs
+++ b/bindings/csharp/Regorus/Regorus.cs
@@ -4,11 +4,20 @@
 using System;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Collections.Generic;
+using System.Text.Json;
 
 
 #nullable enable
 namespace Regorus
 {
+    /// <summary>
+    /// Delegate for callback functions that can be invoked from Rego policies
+    /// </summary>
+    /// <param name="payload">Deserialized JSON object containing the payload from Rego</param>
+    /// <returns>Object that will be serialized to JSON and converted to a Rego value</returns>
+    public delegate object RegoCallback(object payload);
+    
     public unsafe sealed class Engine : System.IDisposable
     {
         private Regorus.Internal.RegorusEngine* E;
@@ -16,6 +25,20 @@ namespace Regorus
         // _isDisposed == 0 means Dispose(bool) has not been called yet.
         // _isDisposed == 1 means Dispose(bool) has been already called.
         private int isDisposed;
+        
+        // Store callback delegates to prevent garbage collection
+        private readonly Dictionary<string, (Internal.RegorusCallbackDelegate Delegate, GCHandle Handle)> callbackDelegates 
+            = new Dictionary<string, (Internal.RegorusCallbackDelegate, GCHandle)>();
+        
+        // Store user callbacks
+        private readonly Dictionary<string, RegoCallback> callbacks = new Dictionary<string, RegoCallback>();
+
+        // JSON serialization options
+        private static readonly JsonSerializerOptions JsonOptions = new JsonSerializerOptions
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            WriteIndented = false
+        };
 
         public Engine()
         {
@@ -51,7 +74,11 @@ namespace Regorus
                 // and unmanaged resources.
                 if (disposing)
                 {
-                    // No managed resource to dispose.
+                    // Unregister all callbacks
+                    foreach (var name in new List<string>(callbackDelegates.Keys))
+                    {
+                        UnregisterCallback(name);
+                    }
                 }
 
                 // Call the appropriate methods to clean up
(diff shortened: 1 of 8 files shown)


kusha:
@microsoft-github-policy-service agree company="Microsoft"

bindings/csharp/Regorus/Regorus.cs
```diff
@@ -4,18 +4,41 @@
...
 
 #nullable enable
 namespace Regorus
 {
+    /// <summary>
+    /// Delegate for callback functions that can be invoked from Rego policies
+    /// </summary>
+    /// <param name="payload">Deserialized JSON object containing the payload from Rego</param>
+    /// <returns>Object that will be serialized to JSON and converted to a Rego value</returns>
+    public delegate object RegoCallback(object payload);
```
anakrish, line 19: Note: An expectation on the Rust side is that an Engine is efficient to clone. It is much faster to add policies to an engine and clone it many times than creating many instances of the engine and adding policies to them individually. 

Cloning an engine will also clone the extension. And a clone could be used from another thread. This means that the extension must be stateless or must be safe to be called from multiple threads.

bindings/csharp/Regorus/Regorus.cs
```diff
@@ -202,7 +229,139 @@ public void SetGatherPrints(bool enable)
             return CheckAndDropResult(Regorus.Internal.API.regorus_engine_take_prints(E));
         }
 
-
+        // Generic callback handler that routes to the appropriate user-provided callback
+        private static unsafe byte* CallbackHandler(byte* payloadPtr, void* contextPtr)
```
anakrish: We could generate closures that forward to RegoCallback delegates and register the closures with Regorus. We might be able to avoid maintaining a dictionary for dispatching.

```suggestion
       private static RegorusCallbackDelegate GenerateRegorusCallback(RegoCallback extension)
       {
           return delegate(byte* payloadPtr, void* contextPtr
           {
               ...
               var result = extension(payloadObject); 
               ...
           };
       }
        private static unsafe byte* CallbackHandler(byte* payloadPtr, void* contextPtr)
```

bindings/ffi/src/lib.rs
```diff
@@ -15,6 +19,60 @@ pub enum RegorusStatus {
     RegorusStatusError,
 }
 
+// Define callback function type for orchestration to Regorus communication
+pub type RegorusCallbackFn = extern "C" fn(payload: *const c_char, context: *mut c_void) -> *mut c_char;
```
anakrish: The `context` is a nice idea and can be leveraged for C and C++ extensions too.

bindings/ffi/src/lib.rs
```diff
@@ -105,13 +163,69 @@ pub extern "C" fn regorus_result_drop(r: RegorusResult) {
...
+}
 
 #[no_mangle]
 /// Construct a new Engine
 ///
 /// See https://docs.rs/regorus/latest/regorus/struct.Engine.html
 pub extern "C" fn regorus_engine_new() -> *mut RegorusEngine {
-    let engine = ::regorus::Engine::new();
+    let mut engine = ::regorus::Engine::new();
+    let _ = engine.add_extension("invoke".to_string(), 2, Box::new(invoke_callback));
```
anakrish: To enable writing `function(args)` instead of `invoke("function", args)` in the rego policy, we could change the behavior of the interpreter so that it looks at the table of foreign language extensions if no function exists for a given name.

On the other hand, `invoke` does make it clear to the reader of the policy that a foreign language function is being used.

anakrish (commented):
@kusha, thank you for taking on this super useful feature has has been in the back burner for a while.

anakrish:
@mkulke I think you will find this interesting.

bindings/ffi/src/lib.rs
kusha: I think `invoke()` would be a cleaner option. Although we can check that it is possible to to trigger invoke from another extension, so that user can optionally use `add_extension("function"...)` that will `invoke("function"...)`

bindings/ffi/src/lib.rs
```diff
@@ -15,6 +16,30 @@ pub enum RegorusStatus {
     RegorusStatusError,
 }
 
+// Store a callback function and its context
+#[no_mangle]
+pub extern "C" fn regorus_register_callback(
```
anakrish, line 21:
```suggestion
pub extern "C" fn regorus_register_invoke_callback(
```

bindings/ffi/src/lib.rs
```diff
@@ -15,6 +16,30 @@ pub enum RegorusStatus {
...
+    if regorus::extensions::invoke::register_callback(name, callback, context) {
+        RegorusStatus::RegorusStatusOk
+    } else {
+        RegorusStatus::RegorusStatusError
+    }
+}
+
+// Remove a callback function
+#[no_mangle]
+pub extern "C" fn regorus_unregister_callback(name: *const c_char) -> RegorusStatus {
```
anakrish, line 35:
```suggestion
pub extern "C" fn regorus_unregister_invoke_callback(name: *const c_char) -> RegorusStatus {
```

bindings/csharp/scripts/invoke.csx
```diff
@@ -0,0 +1,49 @@
+#r "../Regorus/bin/Debug/netstandard2.1/Regorus.dll"
+// No direct reference to regorus_ffi.dll as it's a native DLL
+#r "nuget: Newtonsoft.Json, 13.0.2"
+#r "nuget: System.Data.Common, 4.3.0"
+
+// Create a new engine
+var engine = new Regorus.Engine();
+
+// Enable the invoke extension
+bool enableResult = engine.EnableInvoke();
```
anakrish, line 10: To catch regressions, I'd prefer a test over a script.

anakrish (commented):
Overall looks good to me.

Two comments:
- It would be really great if there was a way to use the `invoke` builtin-extension to implement missing builtins. For example, the [JWT functions](https://www.openpolicyagent.org/docs/latest/policy-reference/#tokens) and the [crypto functions](https://www.openpolicyagent.org/docs/latest/policy-reference/#crypto) would ideally be implemented via `invoke` to route to regulation compliant libraries in the client language (e.g. C#). This doesn't have to be part of the PR.

- It would be great if you can document your usecase (leaving out confidential info) as part of the PR.

kusha:
Callbacks are dropped in favor of host-await interruption in compiled policies.

PR #397, opened 2025-04-16
Create C# sample test suite

anthony-c-martin:


.github/workflows/test-csharp.yml
```diff
@@ -121,17 +121,12 @@ jobs:
...
-
-      - name: Run Test App 8.0
-        run: dotnet run --framework net8.0
-        working-directory: ./bindings/csharp/TestApp
-  
\ No newline at end of file
+          path: ./bindings/csharp/Regorus.Tests/regorus-nuget/
+
+      - name: Restore Regorus.Tests
+        run: dotnet restore /p:RestoreAdditionalProjectSources=./regorus-nuget
```
anakrish, line 127: For my own understanding, why use RestoreAdditionalProjectSources instead of RestoreSources?
The documentation does not explain the rationale: https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets

bindings/csharp/Regorus.Tests/Regorus.Tests.csproj
```diff
@@ -0,0 +1,22 @@
...
+  <ItemGroup>
+    <None Include="../../../tests/**/*.*" Link="tests/%(RecursiveDir)%(Filename)%(Extension)" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MSTest" Version="3.8.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Regorus" Version="0.5.0"/>
```
anakrish, line 20: If the `Version` is not kept up to date, would the test end up picking an older version from somewhere (assuming that we publish to nuget.org)?

bindings/csharp/Regorus.Tests/RegorusTests.cs
```diff
@@ -0,0 +1,176 @@
...
+namespace Regorus.Tests;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Text.Json.Nodes;
+
+[TestClass]
+public class RegorusTests
+{
+    [TestMethod]
+    public void Basic_evaluation_succeeds()
```
anakrish, line 13: Interesting naming convention. Is this the standard for naming tests?

bindings/csharp/Regorus.Tests/RegorusTests.cs
```diff
@@ -0,0 +1,176 @@
...
+      "key": "/run/gcs/c/container0/rootfs",
+      "name": "overlayTargets",
+      "value": true
+    }
+  ]
+}
+""";
+
+        Assert.IsTrue(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result!)), $"Actual: {result}");
+    }
```
anakrish, line 175: Is the formatting off for the line above? It renders weird on github.

anakrish (commented):
@anthony-c-martin, Thanks for the contribution. Overall LGTM. Just a few questions/comments.

.github/workflows/test-csharp.yml
anthony-c-martin: It's my first time encountering these settings, but my understanding is that `RestoreSources` overwrites the list of sources, whereas `RestoreAdditionalProjectSources` adds additional sources to the list.

By default, dotnet will point to the public nuget feed (nuget.org), so if you replace the source list with a local file path, then you'll no longer be able to pull public packages.

Usually, this stuff is all configured through the NuGet.config file that you check in, which is why I've not encountered this before.

bindings/csharp/Regorus.Tests/RegorusTests.cs
anthony-c-martin: I think GitHub is just wrapping words - here's what I see:
image

bindings/csharp/Regorus.Tests/RegorusTests.cs
anthony-c-martin: I haven't really seen any standards around naming for C# tests. This is the convention I tend to follow + recommend - some examples from Bicep:

* https://github.com/Azure/bicep/blob/0b36c48266652fdae833fd171ff220dd9671d594/src/Bicep.Core.IntegrationTests/LambdaTests.cs#L87
* https://github.com/Azure/bicep/blob/0b36c48266652fdae833fd171ff220dd9671d594/src/Bicep.Core.IntegrationTests/LambdaTests.cs#L56
* https://github.com/Azure/bicep/blob/0b36c48266652fdae833fd171ff220dd9671d594/src/Bicep.Core.IntegrationTests/LambdaTests.cs#L227

bindings/csharp/Regorus.Tests/Regorus.Tests.csproj
anakrish: I will follow up on this later.

anakrish (approved):
@anthony-c-martin Thanks for the contribution!

PR #403, opened 2025-04-25
fix: Disallow else blocks for set rules

anakrish:
else blocks following contains and old-style sets will raise a parse error. Consistent with OPA.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1617,6 +1617,12 @@ impl<'source> Parser<'source> {
         let bodies = self.parse_rule_bodies()?;
         span.end = self.end;
 
+        if let RuleHead::Set { .. } = &head {
+            if let Some(RuleBody { span, .. }) = bodies.get(1) {
+                bail!(span.error("else cannot be used with set rules"));
+            }
+        }
+
         if self.rego_v1 && bodies.is_empty() {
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -1617,6 +1617,12 @@ impl<'source> Parser<'source> {
         let bodies = self.parse_rule_bodies()?;
         span.end = self.end;
 
+        if let RuleHead::Set { .. } = &head {
+            if let Some(RuleBody { span, .. }) = bodies.get(1) {
+                bail!(span.error("else cannot be used with set rules"));
```
cemheren, line 1622: these errors need some sort of code so we can catch the code on the server side and return localized error messages (if it's a string it makes it easy to break that contract)

cemheren (approved):
approved with comment

src/parser.rs
anakrish: Thanks for the review. 
I agree about internationalization. There is an issue to revamp error handling ([Avoid use of anyhow](https://github.com/microsoft/regorus/issues/161)) to raise structured errors instead of strings.
That is an opportunity to introduce internationalization too.
Currently, the issue is in backlog but we can bump up the priority if needed.

PR #408, opened 2025-05-13
Basic benchmarking setup with Criterion

unexge:
This PR adds basic benchmarking setup using [Criterion.rs](https://github.com/bheisler/criterion.rs). This would be useful to do performance analyses, and also to measure potential performance optimizations. 

Main motivation for this PR was to understand how Regorus handles constants vs. data. @balcanuc realized that constants are re-evaluated each time and that was causing noticeable performance differences on sets with ~100 elements compared to ~3000 elements. This seems to be only happen with constants and not with data:
lines
We're trying to get more data and plan to create an issue to discuss in more details.

---

The benchmarks can be run with `cargo bench`. Sample output:
```
simple equality check with constant
                        time:   [1.9276 µs 1.9343 µs 1.9417 µs]

simple equality check with data
                        time:   [2.2182 µs 2.2262 µs 2.2344 µs]

allow with simple membership/with constant/32
                        time:   [5.6988 µs 5.7082 µs 5.7188 µs]

allow with simple membership/with data/32
                        time:   [2.2931 µs 2.3194 µs 2.3555 µs]

allow with simple membership/with constant/64
                        time:   [9.0032 µs 9.0216 µs 9.0388 µs]

allow with simple membership/with data/64
                        time:   [2.3670 µs 2.3767 µs 2.3864 µs]

allow with simple membership/with constant/128
                        time:   [17.455 µs 17.558 µs 17.666 µs]

allow with simple membership/with data/128
                        time:   [2.4630 µs 2.4665 µs 2.4701 µs]

allow with simple membership/with constant/512
                        time:   [79.154 µs 79.484 µs 79.844 µs]

allow with simple membership/with data/512
                        time:   [3.2336 µs 3.2544 µs 3.2933 µs]

allow with simple membership/with constant/1024
                        time:   [174.71 µs 174.99 µs 175.29 µs]

allow with simple membership/with data/1024
                        time:   [4.2587 µs 4.2663 µs 4.2744 µs]

allow with simple membership/with constant/2048
                        time:   [350.20 µs 355.25 µs 362.29 µs]

allow with simple membership/with data/2048
                        time:   [6.3407 µs 6.3947 µs 6.4682 µs]

```

Criterion also generates HTML reports and graphs automatically, which can be found at `./target/criterion/report/index.html`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,6 +41,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstream"
 version = "0.6.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -158,6 +164,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5ce89b21cab1437276d2650d57e971f9d548a2d9037cc231abdc0562b97498ce"
 
 [[package]]
+name = "cast"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
+
+[[package]]
 name = "cc"
 version = "1.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


anakrish:
@unexge @balcanuc Thanks for the benchmark! Looking forward to your findings and patterns of writing policy.

Regorus doesn't do many possible optimizations yet (e.g. hoist common sub-expressions out of loops, fold-constants). The examples you provide will be a good motivator for prioritizing what to implement first.

anakrish (approved):

anakrish:
@unexge I took an initial attempt at implementing constant folding in https://github.com/microsoft/regorus/pull/410.

The idea is that expressions that don't depend on input or non-idempotent builtin functions will always produce the same value and can thus be pre-evaluated and cached. The prototype only caches rule values.

image

PR #414, opened 2025-05-27
feat!: Indexes for nodes in the AST

anakrish:
Indexes allow associating extra data with nodes in the AST
using an array and then quickly looking up the array to fetch
the extra data.

- Index eidx for expressions
- Index sidx for statements
- Index qidx for queries.

AST nodes are not cloneable. Therefore once a module is created,
it is not possible to accidentally create two nodes with the same
index inadvertently via clone.

Also added IndexChecker in debug builds. When a module is parsed,
 it will assert that indexes have been constructed correctly.

 AST Cleanup
- Make literal expressions (null, val, number, string etc) also structs
   to match all other expressions
- Merge True and False nodes into a single Bool node.

Also update dependencies.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -106,29 +106,60 @@ pub type Ref<T> = NodeRef<T>;
 #[cfg_attr(feature = "ast", derive(serde::Serialize))]
 pub enum Expr {
     // Simple items that only have a span as content.
-    String((Span, Value)),
-    RawString((Span, Value)),
-    Number((Span, Value)),
-    True(Span),
-    False(Span),
-    Null(Span),
-    Var((Span, Value)),
+    String {
+        span: Span,
+        value: Value,
+        eidx: u32,
+    },
+
+    RawString {
+        span: Span,
+        value: Value,
+        eidx: u32,
+    },
+
+    Number {
+        span: Span,
+        value: Value,
+        eidx: u32,
+    },
+
+    Bool {
+        span: Span,
+        value: Value,
+        eidx: u32,
+    },
+
+    Null {
+        span: Span,
+        value: Value,
+        eidx: u32,
+    },
+
+    Var {
+        span: Span,
+        value: Value,
+        eidx: u32,
+    },
 
     // array
     Array {
         span: Span,
         items: Vec<Ref<Expr>>,
+        eidx: u32,
     },
 
     // set
     Set {
         span: Span,
         items: Vec<Ref<Expr>>,
+        eidx: u32,
     },
 
     Object {
         span: Span,
         fields: Vec<(Span, Ref<Expr>, Ref<Expr>)>,
+        eidx: u32,
     },
 
     // Comprehensions
@@ -136,12 +167,14 @@ pub enum Expr {
         span: Span,
         term: Ref<Expr>,
         query: Ref<Query>,
+        eidx: u32,
     },
 
     SetCompr {
         span: Span,
         term: Ref<Expr>,
         query: Ref<Query>,
+        eidx: u32,
     },
 
     ObjectCompr {
@@ -149,17 +182,20 @@ pub enum Expr {
         key: Ref<Expr>,
         value: Ref<Expr>,
         query: Ref<Query>,
+        eidx: u32,
     },
 
     Call {
         span: Span,
         fcn: Ref<Expr>,
         params: Vec<Ref<Expr>>,
+        eidx: u32,
     },
 
     UnaryExpr {
         span: Span,
         expr: Ref<Expr>,
+        eidx: u32,
     },
 
     // ref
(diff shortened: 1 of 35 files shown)


src/ast.rs
```diff
@@ -243,6 +292,35 @@ impl Expr {
...
+
+    pub fn eidx(&self) -> u32 {
+        use Expr::*;
+        match self {
+            String { eidx, .. }
+            | RawString { eidx, .. }
+            | Number { eidx, .. }
+            | Bool { eidx, .. }
+            | Null { eidx, .. }
+            | Var { eidx, .. }
```
cemheren, line 304: >Var { eid [](http://example.com/codeflow?start=14&length=9)

very nit: can you order this and the fn above the same way? Just a few lines are out of order var bool null

cemheren:
> pub struct Interpreter {

unrelated to this PR, are there some good ways to split this struct up? 3000 line files are a bit hard to follow logically 

---
Refers to: src/interpreter.rs:40 in 46b9f92. [](commit_id = 46b9f929f77dbc959b3c9d97723d477b6147ca60, deletion_comment = False)

src/parser.rs
```diff
@@ -37,9 +44,30 @@ impl<'source> Parser<'source> {
...
+            sidx: 0,
+            qidx: 0,
         })
     }
 
+    fn next_eidx(&mut self) -> u32 {
+        let eidx = self.eidx;
+        self.eidx += 1;
+        eidx
+    }
```
cemheren, line 57: what's the impact of making these increments function calls in rust? are they inlined by compiler usually?

tests/parser/mod.rs
```diff
@@ -102,22 +102,25 @@ fn match_expr_impl(e: &Expr, v: &Value) -> Result<()> {
...
-        Expr::Number(s) => match_span(&s.0, &v["number"]),
-        Expr::True(s) => match_span(s, v),
-        Expr::False(s) => match_span(s, v),
-        Expr::Null(s) => match_span(s, v),
-        Expr::Var(s) => match_span(&s.0, &v["var"]),
-        Expr::Array { span, items } => match_vec(span, items, &v["array"]),
-        Expr::Set { span, items } => match_vec(span, items, &v["set"]),
-        Expr::Object { span, fields } => match_object(span, fields, &v["object"]),
-        Expr::ArrayCompr { span, term, query } => {
+        Expr::String { span: s, .. } => match_span(s, &v["string"]),
```
cemheren, line 106: shouldn't we add a test that checks the new parser indexes are correct?

anakrish: All the parser tests have been updated to also lock down indexes. Additionally, an `IndexChecker` module has been introduced that will assert that the parser has produce continuous, unique indexes starting at zero. `IndexChecker` is enabled in debug builds and is thus run on both Regorus and OPA test suites.

cemheren:
Approved

src/builtins/opa.rs
```diff
@@ -22,10 +22,10 @@ fn opa_runtime(span: &Span, params: &[Ref<Expr>], args: &[Value], _strict: bool)
     ensure_args_count(span, name, params, args, 0)?;
     let mut obj = BTreeMap::new();
 
-    obj.insert(
+    /*obj.insert(
```
dekomissMSFT: >obj [](http://example.com/codeflow?start=6&length=3)

is commenting this out intentional?

tests/parser/mod.rs
```diff
@@ -200,22 +220,41 @@ fn match_expr_impl(e: &Expr, v: &Value) -> Result<()> {
...
 fn match_expr(expr: &Expr, v: &Value) -> Result<()> {
-    match match_expr_impl(expr, v) {
+    /*match match_expr_impl(expr, v) {
         Ok(()) => Ok(()),
         Err(e) => bail!(
             "{e}\nexpr = {expr:#?}\nv={}\n-----------------------\n",
             serde_json::to_string_pretty(v)?
         ),
-    }
+    }*/
```
dekomissMSFT: same here, should this be uncommented or deleted?

dekomissMSFT (approved):

dekomissMSFT:
nit for future PRs: IMO it is much easier to review new iterations of the PR if the new changes are pushed as separate commits. Is there a benefit to force pushing instead?

src/builtins/opa.rs
anakrish: Good catch. This needn't be part of this PR.

tests/parser/mod.rs
anakrish: Good point. The commented-out code allows printing diff at each nesting level. I have changed it so that uncommenting the code will bring it back.
It's mainly for debugging test failures.

anakrish:
> nit for future PRs: IMO it is much easier to review new iterations of the PR if the new changes are pushed as separate commits. Is there a benefit to force pushing instead?

Good point. I have been force-pushing out of habit. I am not really sure anymore why that pattern was following in previous projects. I agree that it is easier to review if commits aren't force pushed.

dekomissMSFT (dismissed):

dekomissMSFT (approved):

PR #422, opened 2025-06-30
implement `net.cidr_is_valid` builtin

tjons:
@anakrish I've been working on implementing the `net.cidr_is_valid` builtin. It currently is complete for IPv4, and so I wanted to try to get some early feedback while I finish support for IPv6.

I'm still quite new to Rust, and so please feel free to give any pointers or suggestions you think I might be missing :crab:

### DIFF
--- a/src/builtins/net.rs
+++ b/src/builtins/net.rs
@@ -1,3 +1,4 @@
+use core::net::IpAddr;
 use std::sync::Arc;
 
 use crate::ast::{Expr, Ref};
@@ -6,7 +7,7 @@ use crate::builtins::utils::ensure_args_count;
 use crate::lexer::Span;
 use crate::value::Value;
 
-use anyhow::{bail, Ok, Result};
+use anyhow::{bail, Ok as aOk, Result};
 
 use super::utils::ensure_string;
 
@@ -14,11 +15,10 @@ pub fn register(m: &mut builtins::BuiltinsMap<&'static str, builtins::BuiltinFcn
     m.insert("net.cidr_is_valid", (cidr_is_valid, 1));
 }
 
-/// Checks if a CIDR string is valid or invalid. Based on the
-/// golang standard library implementation, as that is how the
-/// built-in is implemented in the Open Policy Agent rego standardt
-/// lib.
-/// https://github.com/golang/go/blob/master/src/net/ip.go#L550
+/// Checks if a CIDR string is valid or invalid. Uses the
+/// `net::IpAddr` type to determine if the string is a valid IP,
+/// and checks to ensure that the mask is in bounds for the parsed
+/// IP address type (v4 or v6).
 pub fn cidr_is_valid(
     span: &Span,
     params: &[Ref<Expr>],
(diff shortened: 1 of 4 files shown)


tjons:

src/builtins/net.rs
```diff
@@ -0,0 +1,133 @@
...
+                        break;
+                    },
+                    _ => continue,
+                }
+            }
+        }
+    };
+
+    if v4 {
+        let retval = parse_ipv4(addr_piece.unwrap())?;
```
anakrish: It would be good to avoid `unwrap` since it can panic if the value is None.

One way to avoid `unwrap` would be to do the following early in the function:
```rust
let addr = addr_piece.ok_or_else(||anyhow!(span.error("cidr not valid"))?;
```

src/builtins/net.rs
```diff
@@ -0,0 +1,133 @@
...
+    } else if v6 {
+       let retval = parse_ipv6(addr_piece.unwrap());
+    }
```
anakrish: Need to handle the case when both `v4` and  `v6` are false.

src/builtins/net.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    match mask_piece {
+        None => bail!("cidr not valid"),
```
anakrish: `bail!(span.error(...))` will report the location in the policy where the error happened.

src/builtins/net.rs
```diff
@@ -0,0 +1,133 @@
...
+fn parse_ipv4_or_err(addr: &str) -> Result<[u32; 4]> {
+    // store each octet in the array
+    let mut fields: [u32; 4] = [0; 4];
+    let mut cur_octet: u32 = 0;
+    let mut octet_digits = 0;
+    let mut pos = 0;
+    let mut prev: char = '\0';
+
+    for c in addr.chars() {
+        if c >= '0' && c <= '9' {
```
anakrish: You could do 
``rust
if let Some(d) = c.to_digit(10) {
   ...
} else if c == '.' {
} else {
} 
```
We want to ensure that production code does not use unwrap. Even though the above instance is safe now, it could turn out to be not logically safe in the future.

**Review by anakrish — COMMENTED** [review]
Hey @tjons,
Thanks for the contribution. Overall looks good to me. I've provided some suggestions.

File: src/builtins/net.rs [hunk 5]
```diff
@@ -0,0 +1,133 @@
...
+    m.insert("net.cidr_is_valid", (cidr_is_valid, 1));
+}
+
+pub fn cidr_is_valid(span: &Span, params: &[Ref<Expr>], args: &[Value], _strict: bool) -> Result<Value> {
+    ensure_args_count(span, "cidr_is_valid", params, args, 1)?;
+    let cidr = ensure_string("cidr_is_valid", &params[0], &args[0])?;
+    _cidr_is_valid(cidr)
+}
+
+fn _cidr_is_valid(cidr: Arc<str>) -> Result<Value> {
```
unexge: Maybe we can just use [std::net::IpAddr](https://doc.rust-lang.org/std/net/enum.IpAddr.html) to parse ip address part after removing prefix len. Something like:
```rust
fn is_valid_cidr(cidr: &str) -> Result<bool> {
    let Some((ip_addr, prefix_len)) = cidr.split_once("/") else { bail!("invalid cidr") };
    match ip_addr.parse::<IpAddr>() {
        Ok(_) => Ok(true),
        Err(err) => bail!("invalid cidr")
    }
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ba70f8bc541a186562f38b5c6159febd

tjons: good thinking... I had opted against this because I wanted to make sure that the `regorus` implementation had precise parity with the original Rego built-in, which uses Go's standard library functionality. After reviewing the Rust stdlib implementation you recommended, I see that it's simply implementing [RFC 6943](https://datatracker.ietf.org/doc/html/rfc6943#section-3.1.1), which is probably enough. Go is not strictly compliant with this RFC, which means the Rego built-in is not either. Go's implementation has a couple little quirks but I'm not confident they're worth preserving in Regorus. @anakrish do you care for precise implementation parity with the Rego built-in?

tjons:
@anakrish thanks for the feedback! I'll update this and finish ipv6 this week. Do let me know your thoughts on https://github.com/microsoft/regorus/pull/422#discussion_r2197652414, which will either dramatically simplify the implementation or take a bit longer :)

src/builtins/net.rs
anakrish: @tjons, Thanks for the investigation. I agree - it is not worth preserving parity in this case.

tjons:
@anakrish sorry for the wait, addressed all the feedback! Let me know if anything else is needed for this builtin, I'll get started on the rest this week.

src/builtins/net.rs
```diff
@@ -0,0 +1,88 @@
...
+pub fn cidr_is_valid(
+    span: &Span,
+    params: &[Ref<Expr>],
+    args: &[Value],
+    _strict: bool,
+) -> Result<Value> {
+    ensure_args_count(span, "cidr_is_valid", params, args, 1)?;
+    let cidr = ensure_string("cidr_is_valid", &params[0], &args[0])?;
+
+    match is_valid_cidr(cidr) {
```
anakrish: In order to propagate the error from is_valid_cidr:

```suggestion
    is_valid_cidr(cidr).map_err(|e| span.error(&format!("{e}")))
```

src/builtins/net.rs
```diff
@@ -0,0 +1,88 @@
+use core::net::IpAddr;
+use std::sync::Arc;
+
+use crate::ast::{Expr, Ref};
+use crate::builtins;
+use crate::builtins::utils::ensure_args_count;
+use crate::lexer::Span;
+use crate::value::Value;
+
+use anyhow::{bail, Ok as aOk, Result};
```
anakrish: Do we need to import Ok from anyhow?

src/builtins/net.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+fn is_valid_cidr(cidr: Arc<str>) -> Result<Value> {
+    let Some((ip_addr, prefix_len)) = cidr.split_once("/") else {
+        bail!("invalid CIDR")
+    };
+    match ip_addr.parse::<IpAddr>() {
+        Ok(addr) => {
+            let mask = prefix_len.parse::<i16>()?;
```
anakrish: Might be useful to make the error more informative.
```suggestion
            let mask = prefix_len.parse::<i16>().map_err(|_| anyhow("failed to parse prefix length"))?
```

src/builtins/net.rs
tjons: Nope, this is unnecessary now that we are using `map_err` instead of matching on the `err`

src/builtins/net.rs
tjons: Good idea! Did this for a few other errors as well

anakrish:
Hey @tjons, Overall looks good to me.

There are a bunch of format string related failures. https://github.com/microsoft/regorus/actions/runs/16734832282/job/47424635159?pr=422

Can you also add `v0/netcidrisvalid` and `v1/netcidrisvalid` to https://github.com/microsoft/regorus/blob/main/tests/opa.passing to ensure that tests from OPA testsuite are run?

anakrish:
@tjons No worries. I truly glad that you are contributing.

PR #423, opened 2025-07-02
build: Support manually generating C# bindings via Github action and add a README

dekomissMSFT:
Update the Github action that generates C# bindings for Regorus to support manual generation. This allows for easily generating Nuget packages for any branch, which is useful for integration testing of in-progress Regorus changes with other projects during development.

Also update the Readme for the C# bindings to explain how to generate a nuget package.


.github/workflows/test-csharp.yml
```diff
@@ -63,6 +64,14 @@ jobs:
 
       - run: echo '${{ steps.stepid.outputs.dotnet-version }}'
 
+      - name: Set VersionSuffix environment variable
+        if: github.event_name == 'workflow_dispatch' && startswith(matrix.os, 'ubuntu')
+        run: echo "VersionSuffix=manualtrigger" >> $GITHUB_ENV'
```
anakrish: There is an unmatched '

.github/workflows/test-csharp.yml
```diff
@@ -63,6 +64,14 @@ jobs:
...
+
+      - name: Set VersionSuffix environment variable
+        if: github.event_name == 'workflow_dispatch' && startswith(matrix.os, 'windows')
```
anakrish: Is this needed? `The enclosing job build_nuget` runs only on 'ubuntu_latest'

dekomissMSFT: Yeah, it is needed for the `test_nuget` job since it runs on both windows/ubuntu. I originally was going to set both again inside that job, but found that env variables seem to be preserved across steps so I put them both here. If you think it makes more sense I can still add them to the `test_nuget` job.

anakrish: Would something like the following work:
```yaml
env:
  VersionSuffix: ${{ github.event_name == "workflow_dispatch" && 'manual_trigger'  || null }}
```

dekomissMSFT: Yeah that works, thanks!

anakrish:
@dekomissMSFT The PR looks good. Please squash your commits and fix the commit message.

I've been also following the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) approach of tagging the commit message. Then the commits will be grouped automatically in the release description: https://github.com/microsoft/regorus/releases

anakrish (approved):
LGTM

PR #425, opened 2025-07-03
Implement methods to extract the package name and parameters defined within a policy from the engine

dekomissMSFT:
This PR adds methods to the Regorus engine to extract package names and parameters defined within policies that have been added to the engine.

Parameters are defined as any default rule that is prefixed with "parameters" and only contains one segment following that.

Ex.
default parameters.myParameter := 5    => is a parameter
default parameters.myParameter.nested := 5     => is not a parameter
default parameters := {'myParameter': 5}    => is not a parameter

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -907,6 +907,112 @@ impl Engine {
         serde_json::to_string_pretty(&ast).map_err(anyhow::Error::msg)
     }
 
+    /// Get the package names of each policy added to the engine.
+    ///
+    ///
+    /// ```rust
+    /// # use regorus::*;
+    /// # use anyhow::{bail, Result};
+    /// # fn main() -> Result<()> {
+    /// # let mut engine = Engine::new();
+    /// engine.add_policy("test.rego".to_string(), "package test\n x := 1".to_string())?;
+    /// engine.add_policy("test2.rego".to_string(), "package test.multi.segment\n x := 1".to_string())?;
+    ///
+    /// let package_names = engine.get_policy_package_names()?;
+    /// let value = Value::from_json_str(&package_names)?;
+    ///
+    /// assert_eq!(value[0]["package_name"].as_string()?.as_ref(), "test");
+    /// assert_eq!(value[1]["package_name"].as_string()?.as_ref(), "test.multi.segment");
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "ast")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "ast")))]
+    pub fn get_policy_package_names(&self) -> Result<String> {
+        #[derive(Serialize)]
+        struct Policy<'a> {
+            source: &'a Source,
+            version: u32,
+            package_name: String,
+        }
+
+        let mut package_names = vec![];
+        for m in &self.modules {
+            let package_name = Interpreter::get_path_string(&m.package.refr, None)?;
+            package_names.push(Policy {
+                source: &m.package.span.source,
+                version: 1,
+                package_name: package_name,
+            });
+        }
+
+        serde_json::to_string_pretty(&package_names).map_err(anyhow::Error::msg)
+    }
+
+    /// Get the parameters defined in each policy.
+    ///
+    ///
+    /// ```rust
+    /// # use regorus::*;
+    /// # use anyhow::{bail, Result};
+    /// # fn main() -> Result<()> {
+    /// # let mut engine = Engine::new();
+    /// engine.add_policy("test.rego".to_string(), "package test\n x := 1".to_string())?;
+    ///
+    /// let package_names = engine.get_policy_parameters()?;
+    /// let value = Value::from_json_str(&package_names)?;
+    ///
+    /// println!("{}", package_names);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "ast")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "ast")))]
+    pub fn get_policy_parameters(&self) -> Result<String> {
+        #[derive(Serialize)]
+        struct PolicyParameter {
+            name: String,
+            modifiable: bool,
+            required: bool,
+        }
+
+        #[derive(Serialize)]
+        struct Policy<'a> {
+            source: &'a Source,
+            version: u32,
+            parameters: Vec<PolicyParameter>,
+        }
+
+        let mut policies = vec![];
+        for m in &self.modules {
+            let mut parameters = vec![];
+            for rule in &m.policy {
+                if let Rule::Default { refr, .. } = rule.as_ref() {
+                    let path = Parser::get_path_ref_components(refr)?;
+                    let paths: Vec<&str> = path.iter().map(|s| s.text()).collect();
+
+                    if paths.len() == 2 && paths[0] == "parameters".to_string() {
+                        // Todo: Fetch fields other than name from rego metadoc for the parameter
+                        parameters.push(PolicyParameter {
+                            name: paths[1].to_string(),
+                            modifiable: false,
+                            required: false,
+                        })
+                    }
+                }
+            }
+
+            policies.push(Policy {
+                source: &m.package.span.source,
+                version: 1,
+                parameters: parameters,
+            });
+        }
+
+        serde_json::to_string_pretty(&policies).map_err(anyhow::Error::msg)
+    }
+
     fn make_parser<'a>(&self, source: &'a Source) -> Result<Parser<'a>> {
         let mut parser = Parser::new(source)?;
         if self.rego_v1 {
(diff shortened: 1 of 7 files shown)


src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+            package_name: String,
+        }
+
+        let mut package_names = vec![];
+        for m in &self.modules {
+            let package_name = Interpreter::get_path_string(&m.package.refr, None)?;
+            package_names.push(Policy {
+                source: &m.package.span.source,
+                version: 1,
+                package_name: package_name,
```
anakrish:
```suggestion
                package_name,
```

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+    /// let package_names = engine.get_policy_package_names()?;
+    /// let value = Value::from_json_str(&package_names)?;
+    ///
+    /// assert_eq!(value[0]["package_name"].as_string()?.as_ref(), "test");
+    /// assert_eq!(value[1]["package_name"].as_string()?.as_ref(), "test.multi.segment");
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "ast")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "ast")))]
```
anakrish: This functionality does not need to be guarded with the `ast` feature.

However, if this is experimental, we can introduce a new feature to guard it.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+    pub fn get_policy_package_names(&self) -> Result<String> {
```
anakrish: In purely Rust uses, it is better to return strongly typed objects instead of JSON serialized strings.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+        #[derive(Serialize)]
+        struct Policy<'a> {
```
anakrish: This structure can be made public and the return type of this function.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+            source: &'a Source,
+            version: u32,
+            package_name: String,
```
anakrish: It is not clear why `version`, `source` are returned by `get_policy_package_names`.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+        for m in &self.modules {
+            let package_name = Interpreter::get_path_string(&m.package.refr, None)?;
+            package_names.push(Policy {
+                source: &m.package.span.source,
+                version: 1,
+                package_name: package_name,
+            });
+        }
+
+        serde_json::to_string_pretty(&package_names).map_err(anyhow::Error::msg)
```
anakrish: JSON serialization can be done in the ffi boundary instead of in the Rust API.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+    ///
+    ///
+    /// ```rust
+    /// # use regorus::*;
+    /// # use anyhow::{bail, Result};
+    /// # fn main() -> Result<()> {
+    /// # let mut engine = Engine::new();
+    /// engine.add_policy("test.rego".to_string(), "package test\n x := 1".to_string())?;
+    ///
+    /// let package_names = engine.get_policy_parameters()?;
```
anakrish: Needs correct example.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+    /// let value = Value::from_json_str(&package_names)?;
+    ///
+    /// println!("{}", package_names);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "ast")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "ast")))]
```
anakrish: No need of a feature guard.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+    pub fn get_policy_parameters(&self) -> Result<String> {
+        #[derive(Serialize)]
+        struct PolicyParameter {
```
anakrish: These structs need to be made public so that the contract is strongly typed.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+
+        #[derive(Serialize)]
+        struct Policy<'a> {
+            source: &'a Source,
+            version: u32,
+            parameters: Vec<PolicyParameter>,
+        }
+
+        let mut policies = vec![];
+        for m in &self.modules {
```
anakrish, line 990: Few things to think about:
- Detecting invalid parameters. Define what a valid parameter is.
- Parameter meta-data.
- See if we can evaluate the rule named "parameters" and have the interpreter do most of the heavy lifting.

src/engine.rs
```diff
@@ -907,6 +907,112 @@ impl Engine {
...
+            }
+
+            policies.push(Policy {
+                source: &m.package.span.source,
+                version: 1,
+                parameters: parameters,
+            });
+        }
+
+        serde_json::to_string_pretty(&policies).map_err(anyhow::Error::msg)
```
anakrish: JSON serialization is best done in the FFI layer.

src/engine.rs
dekomissMSFT: What do you think about having the C# bindings parse the JSON back into objects vs letting the caller do it?

src/engine.rs
```diff
@@ -23,6 +23,31 @@ pub struct Engine {
...
 }
 
+#[derive(Debug, Clone, Serialize)]
+pub struct PolicyPackageNameDefinition {
+    pub source_file: String,
+    pub package_name: String,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct PolicyParameterDetails {
```
anakrish:
```suggestion
pub struct PolicyParameter {
```

src/engine.rs
```diff
@@ -23,6 +23,31 @@ pub struct Engine {
...
+    pub required: bool,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct PolicyModifierDetails {
+    pub name: String,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct PolicyParameterDefinition {
```
anakrish:
```suggestion
pub struct PolicyParameters {
```

src/engine.rs
```diff
@@ -907,6 +932,109 @@ impl Engine {
...
+    /// engine.add_policy("test2.rego".to_string(), "package test.multi.segment\n x := 1".to_string())?;
+    ///
+    /// let package_names = engine.get_policy_package_names()?;
+    ///
+    /// assert_eq!("test", package_names[0].package_name);
+    /// assert_eq!("test.multi.segment", package_names[1].package_name);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn get_policy_package_names(&self) -> Result<Vec<PolicyPackageNameDefinition>> {
```
anakrish, line 955: This function is quite similar to `get_packages`. Do we really need both?

src/engine.rs
```diff
@@ -907,6 +932,109 @@ impl Engine {
...
+                        // Todo: Fetch fields other than name from rego metadoc for the parameter
+                        parameters.push(PolicyParameterDetails {
+                            name: paths[1].to_string(),
+                            modifiable: false,
+                            required: false,
+                        })
+                    }
+                }
+
+                // Extract modifiers to the parameters from the policy rule
```
anakrish, line 1011: It is not clear to me what a modifier is. Is seems to be similar to a parameter.

src/engine.rs
dekomissMSFT: I missed this method. Really the only difference is that the one I added includes the policy file name along with each package name. Do you think I should consolidate the two since it'd be a breaking change to the other method?

src/engine.rs
dekomissMSFT: It's basically a parameter value. I named it modifier since we may eventually expand the concept to allow you to override any rule within the definition, not just the parameters. Are you okay with leaving the name as is and finalizing it later once we start working on parameters again?

src/engine.rs
anakrish: I think that's a great idea. To unblock this PR, we can attempt it next.

src/engine.rs
anakrish: We can attempt this later.

src/engine.rs
anakrish: Good point. We can hold off on merging the methods for now. Since we are adding functionality under the `azure_policy` feature, we can focus on the overall API later.

src/engine.rs
anakrish: Sounds good!

anakrish:
@dekomissMSFT Looks good to me. Once you squash you commits, I can approve it. Please do use a semantic commit message (e.g.: feat).

anakrish (approved):
LGTM

Issue #426, opened 2025-07-06
Early return doesn't work with 'some' keyword


ZabelinK:
Early return (also known as Early Exit in standard OPA top-down evaluation — see: https://www.openpolicyagent.org/docs/policy-performance#early-exit-in-rule-evaluation) works when using iteration with the placeholder syntax [_], but it does not work with the some keyword, even when the return value is constant.

For example, consider a policy using an int_equals_3 extension that compares integers to 3 and prints each compared value. In this case:

Evaluating the rule early_return_works will print: 1, 2, 3.

Evaluating the rule early_return_doesnt_work will print: 1, 2, 3, 4, 5, 6.

This happens even though the return value of early_return_doesnt_work is constant, and any iterations over values greater than 3 do not affect the final result of the rule.

```
package item_match

import rego.v1

items := [1, 2, 3, 4, 5, 6]

early_return_works if {
    int_equals_3(items[_])
}

early_return_works_doesnt_work if {
    some item in items
    int_equals_3(item)
}
```

<details>
<summary>Full example with output</summary>

```
use anyhow::Result;
use regorus::{Engine, Value};

fn main() -> Result<()> {
    let mut engine = Engine::new();

    // Simple extension that checks if a number equals 3
    engine.add_extension("int_equals_3".to_string(), 1, Box::new(|args: Vec<Value>| {
        let n = args[0].as_i64().unwrap_or(0);
        let res = n == 3;
        println!("{} == 3 -> {}", n, res);
        Ok(Value::Bool(res))
    }))?;

    engine.add_policy(
        "array_match.rego".to_string(),
        r#"
package item_match

import rego.v1

items := [1, 2, 3, 4, 5, 6]

early_return_works if {
    int_equals_3(items[_])
}

early_return_doesnt_work if {
    some item in items
    int_equals_3(item)
}
"#.to_string(),
    )?;

    println!("early_return_works:");
    engine.eval_query("data.item_match.early_return_works".to_string(), false)?;
    
    println!("early_return_doesnt_work:");
    engine.eval_query("data.item_match.early_return_doesnt_work".to_string(), false)?;
    
    Ok(())
} 
```

Output:
```
early_return_works:
1 == 3 -> false
2 == 3 -> false
3 == 3 -> true
early_return_doesnt_work:
1 == 3 -> false
2 == 3 -> false
3 == 3 -> true
4 == 3 -> false
5 == 3 -> false
6 == 3 -> false

```
</details>

PR #427, opened 2025-07-06
Early return for 'some in' statement

ZabelinK:
Fix for the issue described here: https://github.com/microsoft/regorus/issues/426

Introducing early return checks in eval_some_in.

And adjusting kata test output, since there is no more additional checks after rules with 'some in' found a match.


ZabelinK:

ZabelinK:
Sure! Just synced the fork.

PR #441, opened 2025-07-26
fix: set default hash in case git is not available

ma-ts:
Currently, building regorus does not work if `git` is not available in the build context, which is very common in a Docker build.
This falls back to an 'unknown' hash in case git is not available.


anakrish:
Hey @ma-ts, just a gentle ping.

PR #443, opened 2025-07-27
perf: Optimize clone

anakrish:
Introduce the notion of CompiledPolicy to hold stuff that remains immutable during evaluation - e.g. rules, function, schedules etc

Cloning takes about 60 nano seconds for an engine loaded with ACI policies. Earlier it used to take 40 microseconds. Thus there is easily more than 100x speedup.

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -36,14 +36,42 @@ enum FunctionModifier {
     Value(Value),
 }
 
-#[derive(Debug, Clone)]
-pub struct Interpreter {
-    modules: Vec<Ref<Module>>,
-    module: Option<Ref<Module>>,
+#[derive(Debug, Clone, Default)]
+pub struct CompiledPolicy {
+    modules: Rc<Vec<Ref<Module>>>,
     schedule: Option<Schedule>,
+    rules: Map<String, Vec<Ref<Rule>>>,
+    default_rules: Map<String, Vec<DefaultRuleInfo>>,
+    imports: BTreeMap<String, Ref<Expr>>,
+    functions: FunctionTable,
+    rule_paths: Set<String>,
+}
+
+type RuleValues = BTreeMap<Vec<Value>, (Value, Ref<Expr>)>;
+
+#[derive(Debug)]
+pub struct Interpreter {
+    compiled_policy: Rc<CompiledPolicy>,
+
+    data: Value,
+
+    #[cfg(feature = "coverage")]
+    coverage: Map<Source, Vec<bool>>,
+    #[cfg(feature = "coverage")]
+    enable_coverage: bool,
+
+    traces: Option<Vec<Rc<str>>>,
+    strict_builtin_errors: bool,
+
+    gather_prints: bool,
+    prints: Vec<String>,
+
+    extensions: Map<String, (u8, Rc<Box<dyn Extension>>)>,
+
+    module: Option<Ref<Module>>,
     current_module_path: String,
     input: Value,
-    data: Value,
+
     init_data: Value,
     with_document: Value,
     with_functions: BTreeMap<String, FunctionModifier>,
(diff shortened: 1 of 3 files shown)


src/interpreter.rs
```diff
@@ -81,6 +94,47 @@ impl Default for Interpreter {
     }
 }
 
+impl Clone for Interpreter {
+    fn clone(&self) -> Self {
+        Self {
+            compiled_policy: self.compiled_policy.clone(),
```
dekomissMSFT, line 100: >compiled_policy: self.compiled_policy.clone(), [](http://example.com/codeflow?start=12&length=46)

So does the perf improvement primarily come from the fact that cloning an Rc of an object doesn't do a deep copy? Can the Rust compiler guarantee that the data within the CompiledPolicy won't get modified?

src/interpreter.rs
```diff
@@ -2602,7 +2660,7 @@ impl Interpreter {
     }
 
     fn ensure_module_evaluated(&mut self, path: String) -> Result<()> {
-        for module in self.modules.clone() {
+        for module in self.compiled_policy.modules.clone().iter().cloned() {
```
dekomissMSFT, line 2663: >modules.clone() [](http://example.com/codeflow?start=43&length=15)

Why do we clone the modules here if they're part of CompiledPolicy? #Resolved

dekomissMSFT: Oh wait, I see this is also an Rc

src/interpreter.rs
```diff
@@ -2642,7 +2700,7 @@ impl Interpreter {
 
     fn ensure_rule_evaluated(&mut self, path: String) -> Result<()> {
         let mut matched = false;
-        if let Some(rules) = self.rules.get(&path) {
+        if let Some(rules) = self.compiled_policy.rules.get(&path) {
             matched = true;
             for r in rules.clone() {
```
dekomissMSFT, line 2705: > rules.clone() [](http://example.com/codeflow?start=20&length=14)

why do we need to clone the rules still?

src/interpreter.rs
```diff
@@ -81,6 +94,47 @@ impl Default for Interpreter {
...
+
+            extensions: self.extensions.clone(),
+
+            #[cfg(feature = "coverage")]
+            coverage: self.coverage.clone(),
+            #[cfg(feature = "coverage")]
+            enable_coverage: self.enable_coverage,
+
+            // The following fields always get cleared during an evaluation.
+            // Hence, they need not be copied.
```
dekomissMSFT, line 121: is this why you need to implement the Clone function yourself rather than using the #derive(Clone)?

dekomissMSFT: also, when/where do we clone the interpreter?

src/interpreter.rs
anakrish: > So does the perf improvement primarily come from the fact that cloning an Rc of an object doesn't do a deep copy?

Yes. With this change, cloning is close to O(1). Other than `extensions`, no other fields are deep copied. 

>Can the Rust compiler guarantee that the data within the CompiledPolicy won't get modified

In Rust, the language guarantees that there can be only one mutable reference to an objects, even across threads.So, race conditions are avoided. Additionally, to modify an object wrapped by an `Arc`, [make_mut](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) must be called which also avoid race conditions. So, in effect Arc<CompiledPolicy>  can be shared between threads safely, without any locking.

src/interpreter.rs
anakrish: > is this why you need to implement the Clone function yourself rather than using the #derive(Clone)?

Yes.

anakrish: > also, when/where do we clone the interpreter?

Interpreter is held on to by the Engine and gets cloned when the Engine is cloned. The Interpreter's clone will share the CompiledPolicy with the instances from which it was cloned.

Regorus started off with the Interpreter being the core entity and later Engine was added.
We need to refactor and clean up these concepts - the Interpreter could ideally be split up.

dekomissMSFT: ah, I was confused because I couldn't find a call to interpreter.clone, but it happens automatically whenever the Engine is cloned. Makes sense.

src/interpreter.rs
dekomissMSFT: makes sense, thanks for the explanation!

src/interpreter.rs
anakrish: It is due to Rust's ownership rules - "there can be only one mutable reference or many immutable references to an object at a given time".

 In line 2703, when we hold on to `rules`, we are also holding onto an immutable reference to `self.compiled_policy` and also `self`. The call to `self.eval_rule` in line 2708 creates a mutable reference which violates the ownership rules. Therefore, we clone the rules and use it in line 2705, eliminating the immutable reference to `rules`, `self.compiled_policy` and `self`. The clone here is efficient - just a ref count increment.

The reason the ownership rule exists is due to this: Assume that, we don't clone and hold on to `self.compiled_policy.rules`. As we are iterating through the rules, `self.eval_rule` is called which being a mutable method could modify `self.compiled_policy.rules` which could lead to undefined behavior. In other languages, it is easy to create a similar scenario - iterate through a linked list and modify the list during the iteration.

PR #449, opened 2025-08-04
Fix install of regorus_ffi C++ bindings

milianw:
When using a clean build dir the install failed as it referenced a non-existing `regorus_ffiCorrosion.cmake` file. I believe I used the shorter `regorus_ffi` as my `EXPORT` in the `corrosion_install` at some point and then later didn't notice that I referenced a stale generated file when I initially handed in this PR. We must make sure that the same identifier is used here too. See also the documentation from `corrosion_install`:

> * **EXPORT**: Creates an export that can be installed with `install(EXPORT)`. <export-name> must be globally unique.
>             Also creates a file at ${CMAKE_BINARY_DIR}/corrosion/<export-name>Corrosion.cmake that must be included in the installed config file.


anakrish (approved):
LGTM

milianw:
Thanks! I cannot push "merge", will you please do that for me?

PR #452, opened 2025-08-05
feat: Types

anakrish:
Regorus's type system is a first of many features intended to
enable type checking and various other constraints on Rego policies.

The type system is insipired from:
   - JSON schema
   - Bicep

The type system is a subset of JSON schema that are useful
for type-checking policies written for cloud resources.
It is expected to be also useful for most general scenarious as well
since the Rego `input` generally refers to an instance of a type
that can be expressed in many strongly typed programming languages.

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -1,0 +1,1033 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/// There are two type systems of interest:
+///     1. JSON Schema used by Azure Policy for some of its metadata.
+///     2. Bicep's type system generated from Azure API swagger files.
+///        https://github.com/Azure/bicep-types/blob/main/src/Bicep.Types/ConcreteTypes
+///
+/// JSON Schema is standardized and well documented, with good tooling support.
+/// JSON Schema is quite flexible. The following schema:
+/// {
+///   "allOf": [
+///     {
+///       "properties": {
+///         "name": {"type": "string" }
+///       },
+///       "required": ["name"]
+///     },
+///     {
+///       "properties": {
+///         "age": {"type": "integer" }
+///       },
+///       "required": ["age"]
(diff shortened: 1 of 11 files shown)


src/schema.rs
```diff
@@ -0,0 +1,569 @@
...
+                    }
+                }
+                "boolean" => Ok(Type::Boolean),
+                "integer" => Ok(Type::Integer),
+                "null" => Ok(Type::Null),
+                "number" => Ok(Type::Number),
+                "object" => match &schema["properties"] {
+                    Value::Object(props) => {
+                        // Enforce strict schema: only allow "type" and "properties" for object.
+                        // Any extra fields will result in an error to catch schema mistakes
```
dekomissMSFT: Is this meant to parse ARM resource payloads? If so, there are other top level fields we need to account for as well

src/schema.rs
```diff
@@ -0,0 +1,569 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let schema = Value::deserialize(deserializer)?;
+        Type::from_value(schema).map_err(serde::de::Error::custom)
+    }
+}
+
+#[cfg(test)]
+mod test {
```
dekomissMSFT: >test [](http://example.com/codeflow?start=4&length=4)

nit: should tests be in a separate file? what's the convention in rust?

src/schema.rs
anakrish: This is intented as the schema for ARM resource payloads. It is a subset of JSON schema.

src/schema.rs
anakrish: Good suggestion. I've moved it to a separate file.

cemheren (approved):
Approving with a high level skim, I think it's a great start and I like that it's extensible enough if and when we need to add more elements.

PR #456, opened 2025-08-11
Schema registry

anakrish:


src/schema/validate.rs
```diff
@@ -0,0 +1,1001 @@
...
+    *,
+};
+use alloc::collections::BTreeMap;
+use regex::Regex;
+
+type String = Rc<str>;
+
+/// Validation errors that can occur when validating a Value against a Schema.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ValidationError {
```
cemheren: >ValidationError [](http://example.com/codeflow?start=9&length=15)

is this normally the pattern in rust to put multiple entities in one file?  #Resolved

src/schema/validate.rs
```diff
@@ -0,0 +1,1001 @@
...
+            _ => Err(ValidationError::TypeMismatch {
+                expected: "number".into(),
+                actual: Self::value_type_name(value),
+                path: path.into(),
+            }),
+        }
+    }
+
+    fn validate_boolean(value: &Value, path: &str) -> Result<(), ValidationError> {
+        match value {
```
cemheren, line 197: asking to understand: 

Why would be use the match statement here vs passing a concrete type to this method?

benches/schema_validation_benchmark.rs
```diff
@@ -0,0 +1,886 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use regorus::Value;
+use regorus::{Schema, SchemaValidator};
+use serde_json::json;
+
+// Observed: validate_string - 3.19 ns/iter
```
dpokluda, line 6: >3.19 ns/iter [](http://example.com/codeflow?start=31&length=12)

Just curious, does it make sense to include these? Don't they rely on the hardware used? (I don't mind these being here, just curious. One reason to keep them is to get an idea of order of magnitude difference between individual methods.) #Resolved

src/schema/validate.rs
anakrish: In Rust, each file is a module and a module can be specified only in a single file.
Sub modules can be specified in their own separate files within a directory whose name matches the parent module.

I have not seen too much of a standard pattern here. There is no hard rule that each entity must exist in its own file.
It's more about the file size I guess. In this case, we could actually move the ValidationError to its own separate file (module) and refer to it elsewhere.

benches/schema_validation_benchmark.rs
anakrish: Yes, the exact values are subject to hardware used and the load on the system. That is why we don't run these benchmarks in CI (yet) - the environments are too variable.

The comment is there as an indicator of what these numbers can look like.

src/schema/validate.rs
anakrish: Not much difference. If I had written from scratch, I would probably have written it the way you are suggesting - pass a boolean here.

The current version is more for consistency - where each validate_ method takes a Value.
But I don't think the consistency matters that much either.

cemheren: I see if you copied from somewhere can we link to the original? 

I do see value in these types of changes since it makes the reading easier .w.r.t  following which variable goes where. But it probably makes much less difference when actually using an IDE

anakrish: >I see if you copied from somewhere can we link to the original?

I didn't copy. I got CoPilot to write out the code for me, more or less the way I want it.
If I had written out the entire implementation by hand, it would look more or less similar, few stylistic differences here and there, with less comments.

src/schema/validate.rs
```diff
@@ -0,0 +1,1001 @@
...
+
+    fn validate_string(
+        value: &Value,
+        min_length: Option<usize>,
+        max_length: Option<usize>,
+        pattern: Option<&String>,
+        path: &str,
+    ) -> Result<(), ValidationError> {
+        match value {
+            Value::String(s) => {
```
cemheren: >s [](http://example.com/codeflow?start=26&length=1)

nit: better variable name than s please #Resolved

src/schema/validate.rs
```diff
@@ -0,0 +1,1001 @@
...
+    /// ```
+    pub fn validate(value: &Value, schema: &Schema) -> Result<(), ValidationError> {
+        Self::validate_with_path(value, schema, "")
+    }
+
+    /// Internal validation function that tracks the current path for error reporting.
+    fn validate_with_path(
+        value: &Value,
+        schema: &Schema,
+        path: &str,
```
dpokluda, line 324: >path: &str [](http://example.com/codeflow?start=8&length=10)

Curious (don't know Rust): What is this? I don't see it being used.

src/schema/validate.rs
anakrish: I've moved the error type to a separate file (module).

src/schema/validate.rs
anakrish: Fixed.

src/schema/validate.rs
cemheren: thanks that's helpful

src/schema/validate.rs
cemheren: ok looking back, I do think methods should have concrete input types for better readability. match introduces an unnecessary nesting to everything

src/schema/error.rs
```diff
@@ -0,0 +1,276 @@
...
+                actual_length,
+                min_length,
+                max_length,
+                path,
+            } => {
+                let constraint_desc = match (min_length, max_length) {
+                    (Some(min), Some(max)) => format!("between {min} and {max} characters"),
+                    (Some(min), None) => format!("at least {min} characters"),
+                    (None, Some(max)) => format!("at most {max} characters"),
+                    (None, None) => "within valid length".to_string(),
```
cemheren, line 139: > (None, None) => "within valid length".to_string(), [](http://example.com/codeflow?start=19&length=51)

how does this case throw?

cemheren: I mean if there are no length constraints it can't be violated right?

anakrish: Yes. That case can never be reached. However, the compiler is not smart enough to prove it and requires handling the (None, None) case.
Another thing to do here is to add an [`unreachable!'](https://doc.rust-lang.org/std/macro.unreachable.html). But that introduces a call to panic.

src/schema/validate.rs
```diff
@@ -0,0 +1,731 @@
...
+                        if let Some(prop_schema) = properties.get(prop_name_str) {
+                            // Property is defined in schema, validate against it
+                            Self::validate_with_path(prop_value, prop_schema, &make_prop_path())
+                                .map_err(|e| ValidationError::PropertyValidationFailed {
+                                    property: prop_name_str.clone(),
+                                    path: path.into(),
+                                    error: Box::new(e),
+                                })?;
+                        } else if let Some(additional_schema) = additional_properties {
+                            // Property is not defined but additional properties are allowed
```
dekomissMSFT, line 406: >// Property is not defined but additional properties are allowed [](http://example.com/codeflow?start=28&length=64)

nice, was going to ask for this. Can this be specified in any object? I'm thinking of the following use cases:
1. Provide a default schema with no required properties and additional properties allowed (essentially allow any object to be evaluated)
2. Sometimes resource types have freeform object properties where the user can put any key/value. I assume this supports that?

src/schema/validate.rs
```diff
@@ -0,0 +1,731 @@
...
+
+                        if let Some(prop_schema) = properties.get(prop_name_str) {
+                            // Property is defined in schema, validate against it
+                            Self::validate_with_path(prop_value, prop_schema, &make_prop_path())
+                                .map_err(|e| ValidationError::PropertyValidationFailed {
+                                    property: prop_name_str.clone(),
+                                    path: path.into(),
+                                    error: Box::new(e),
+                                })?;
+                        } else if let Some(additional_schema) = additional_properties {
```
dekomissMSFT, line 405: >additional_properties [](http://example.com/codeflow?start=64&length=21)

say the user does create a schema where there's an object with additional properties. How will validation of the rule against that look?

For example, if I have a rule try to access myObject.propertyNotDefinedInTheSchema. Will we allow controlling whether additional properties can be accessed in the rule or not?

src/schema/tests/validate/resource.rs
```diff
@@ -0,0 +1,1328 @@
...
+        "enum": ["Microsoft.Compute/virtualMachines", "Microsoft.Storage/storageAccounts", "Microsoft.Network/virtualNetworks"],
+        "description": "Azure Resource types"
+    });
+
+    let schema = Schema::from_serde_json_value(schema_json).unwrap();
+    Rc::new(schema)
+}
+
+// Helper function to create a virtual machine resource schema
+fn create_vm_resource_schema() -> Rc<Schema> {
```
dekomissMSFT, line 22: >create_vm_resource_schema [](http://example.com/codeflow?start=3&length=25)

do we plan to document the schema of a schema object using something like typespec or swagger? These examples are super helpful for understanding what it looks like, long term I think it would be beneficial to have it formally defined somewhere. It doesn't have to be done in this PR

src/schema/validate.rs
anakrish: > Provide a default schema with no required properties and additional properties allowed (essentially allow any object to be evaluated)
>Sometimes resource types have freeform object properties where the user can put any key/value. I assume this supports that?

Following would be an example of both.

```json
{
  "type": "object",
   "additionalProperties": {
       "type": "any"
   }
}
```

Note that in JSON, unlike Rego, the keys must be strings. But based on the above, it can be any string.

JSON schema is very flexible, the same can be expressed as:
```json
{
  "allOf":[
    {
  		"type": "object"
	},
	{
       "additionalProperties": {
       		"type": "any"
   	   }
    }
  ]
}
```

We don't allow the above syntax since it makes it too difficult to reason about.

src/schema/validate.rs
anakrish: We will validate all input field accesses according to the input's schema.
Initially, we will keep the implementation simple.
```rego
      input.property # will be validated
      input.property.subproperty # will be validated
      
      p = input.property
      p.subproperty # will not be validated. 
```
We will handle all cases once we incorporate type propagation.

> Will we allow controlling whether additional properties can be accessed in the rule or not

If there is a use-case, we can enable this configuation at the target level or at the policy definition level.

src/schema/tests/validate/resource.rs
anakrish: It is already checked in. https://github.com/microsoft/regorus/blob/main/src/schema/meta.schema.json.
Folks that are authoring schemas by hand, can use this meta-schema to help with authoring.

It is likely that we will just pull in all the  resource type schemas from Bicep's generated repository.
E.g: https://github.com/Azure/bicep-types-az/blob/main/generated/containerregistry/microsoft.containerregistry/2025-05-01-preview/types.json

It is in a serialized format different from JSON Schema, but it should be relatively straightforward for us to pull in.

src/schema/error.rs
anakrish: I'm resolving this.
The good thing is that we are using strongly typed errors instead of [anyhow](https://docs.rs/anyhow/1.0.99/anyhow/) that is used in most places in Regorus.

When we attempt https://github.com/microsoft/regorus/issues/161, we can revisit the specific error messages.

PR #471, opened 2025-08-31
Implement `net.cidr_contains` builtin

tjons:
Implement the `net.cidr_contains` builtin in Regorus. Added the [IpNet](https://docs.rs/ipnet/latest/ipnet/index.html) crate to handle and standardize CIDR parsing, which will be useful for the remainder of the `net.*` built-ins.


src/builtins/net.rs
```diff
@@ -59,6 +61,50 @@ fn is_valid_cidr(cidr: Arc<str>) -> bool {
...
+    let cidr_or_ip = ensure_string("cidr_contains", &params[1], &args[1])?;
+    let contains = _cidr_contains(cidr, cidr_or_ip);
+
+    match contains {
+        Ok(r) => Ok(Value::from(r)),
+        // The rego implementation will retur an error in strict mode, see
+        // https://github.com/open-policy-agent/opa/blob/main/v1/test/cases/testdata/v1/netcidrcontains/test-netcidrcontains-0100.yaml
+        // as an example, so we will propagate the error if the builtin is
+        // run in strict mode.
+        Err(_) if strict => bail!(span.error("{e}")),
```
anakrish:
```suggestion
        Err(e) if strict => bail!(span.error(&format!("{e}"))),
```

src/builtins/net.rs
```diff
@@ -79,8 +125,62 @@ mod net_tests {
         for cidr in invalids {
             assert!(
                 !is_valid_cidr(Arc::from(cidr)),
-                "Invalid CIDR {cidr} deemed valid"
+                "Invalid CIDR {cidr} deemed valids"
```
anakrish:
```suggestion
                "Invalid CIDR {cidr} deemed valid"
```

anakrish: The original version read better.

Cargo.toml
```diff
@@ -108,6 +108,7 @@ uuid = { version = "1.15.1", default-features = false, features = ["v4", "fast-r
 jsonschema = { version = "0.30.0", default-features = false, optional = true }
 chrono = { version = "0.4.40", optional = true }
 chrono-tz = { version = "0.10.1", optional = true }
+ipnet = { version = "2.11.0", optional = true }
```
anakrish: Worth exploring `default-features = false`. That will allow use in no_std environments as well.

src/builtins/net.rs
tjons: I fat-fingered this one by accident

Cargo.toml
tjons: @anakrish thanks for the pointer. I'm a little unsure if there's anything more I need to do to ensure this builtin works in `no_std` environments other than adding `default-features = false`. I see that there is an `opa-no-std` featureset defined, but not all of the dependencies marked `default-features = false` are included there. Could you help me understand if there's anything else I should do?

tjons:
@anakrish these java bindings are killing me... I figured out how to update them based on the github action, but I have a couple questions:

1. Is there a reason the Java bindings fail but all other languages succeed? All languages with a `Cargo.lock` file seem to build bindings with `--frozen`, and while the Java binding uses `--locked` as well, `--frozen` should be _more_ restrictive than `--locked`. 
2. Could I contribute (in a future PR) some sort of tooling to check this locally before pushing to CI? Perhaps leveraging https://docs.rs/cargo-task/latest/cargo_task/ or adding something to the `pre-push` hook? Usually in Go projects we have `Makefile` targets to check code generation.

Cargo.toml
anakrish: There is nothing  more that needs to be done other than adding it to the `opa-no-std` list above. 
I see that ipnet has an `std` feature which implies that it has `no_std` support.

anakrish:
Hey @tjons,

1.a I don't exactly know why Java bindings fail. In the past, I have seen it fail due to stricter clippy errors. I thought my recent [PR](https://github.com/microsoft/regorus/blob/85753aaf379c0df6fb4e60a76a1a2650b38dad53/.github/workflows/test-java.yml#L25) that uses the same rust toolchain uniformly would fix it. But looks like it doesn't.
1.b Yes. It should also use --frozen.

2. Yes, testing it locally is a great idea. The current pre commit hook approach is something I put together to prevent me from breaking stuff, while I was learning Rust. If there are better approaches (say cargo-task) we can totally use it.

anakrish (approved):
LGTM. Thanks for the contribution @tjons

PR #491, opened 2025-10-27
build: Add xtask automation for binding version management

anakrish:
Introduces a dedicated xtask crate that keeps language binding versions in sync with the core regorus crate, following the workflow pattern used by rust-analyzer, gitoxide, and ripgrep.

Key features:
- Git-based change detection: compares binding source files against a base ref (merge-base with origin/main by default) plus unstaged/ untracked files to identify which bindings have been modified
- SemVer-aware bumping: binding edits trigger a minor version increment (e.g. 0.5.1 → 0.6.0) under pre-1.0 semantics, signaling potential breaking changes; clean bindings simply align to the root version
- Multi-language support: updates Cargo manifests (Rust FFI, Java, Python, WASM, Ruby), Maven pom.xml (Java), Ruby version constants, and C# project files in a single pass
- CI integration: --check mode fails fast when manifests are out of sync, ensuring pre-commit and release-plz workflows catch stale versions before merge

Integration points:
- release-plz.toml: runs cargo xtask bindings --base-ref origin/main after bumping the root crate, so binding versions are updated atomically during the release process
- scripts/pre-commit: invokes cargo xtask bindings --check to block commits that would leave bindings out of sync
- .cargo/config.toml: defines cargo xtask alias for convenience

Documentation includes inline examples showing how version bumps behave when bindings are ahead/behind the root, and notes that the minor field acts as the major version under SemVer 0.y.z initial development phase.

### DIFF
--- a/.github/workflows/test-java.yml
+++ b/.github/workflows/test-java.yml
@@ -29,12 +29,19 @@ jobs:
           cargo clippy --all-targets --no-deps -- -Dwarnings
           cargo build --release --manifest-path bindings/java/Cargo.toml --locked
 
+      - name: Capture binding version
+        run: |
+          version=$(cargo metadata --manifest-path bindings/java/Cargo.toml --format-version 1 \
+            | jq -r '.packages[] | select(.name == "regorus-java") | .version')
+          echo "REGORUS_JAVA_VERSION=$version" >> "$GITHUB_ENV"
+
       - name: Build jar
         run: mvn package
         working-directory: ./bindings/java
 
       - name: Test jar
         run: |
-          javac -cp target/regorus-java-0.2.2.jar Test.java
-          java -Djava.library.path=target/release -cp target/regorus-java-0.2.2.jar:. Test
+          jar="regorus-java-${REGORUS_JAVA_VERSION}.jar"
+          javac -cp "target/${jar}" Test.java
+          java -Djava.library.path=target/release -cp "target/${jar}:." Test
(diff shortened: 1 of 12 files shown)


bindings/csharp/Benchmarks/Benchmarks.csproj
```diff
@@ -16,7 +16,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Regorus" Version="0.6.0$(RegorusPackageVersionSuffix)"/>
+    <PackageReference Include="Regorus" Version="0.8.0$(RegorusPackageVersionSuffix)"/>
```
dekomissMSFT, line 19: >8.0 [](http://example.com/codeflow?start=51&length=3)

should we bump the version of the C# bindings to 0.8.1 so we can publish a new nuget with your latest changes?

dekomissMSFT (approved):
:shipit:

bindings/csharp/Benchmarks/Benchmarks.csproj
anakrish: This change is just to get the version numbers consistent. I think once this goes in, the xtask should take care of the updates automatically. But I have a feeling we might need to iterate a few times to nail it.

dekomissMSFT: Sounds good :)

Issue #94, opened 2024-01-05
[Lib] Implement token verification builtins


anakrish:
- [ ] [io.jwt.decode](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtdecode)
- [ ] [io.jwt.decode_verify](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtdecode_verify)
- [ ] [io.jwt.verify_es256](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_es256)
- [ ] [io.jwt.verify_es384](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_es384)
- [ ] [io.jwt.verify_es512](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_es512)
- [ ] [io.jwt.verify_hs256](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_hs256)
- [ ] [io.jwt.verify_hs384](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_hs384)
- [ ] [io.jwt.verify_hs512](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_hs512)
- [ ] [io.jwt.verify_ps256](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_ps256)
- [ ] [io.jwt.verify_ps384](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_ps384)
- [ ] [io.jwt.verify_ps512](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_ps512)
- [ ] [io.jwt.verify_rs256](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_rs256)
- [ ] [io.jwt.verify_rs384](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_rs384)
- [ ] [io.jwt.verify_rs512](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtverify_rs512)

[...]

PR #498, opened 2025-11-07
feat: JWT verification builtins (#94)

wetpeanuts:
**Implemented:**
- Implemented builtins: io.jwt.[decode|decode_verify|verify_xxxxx] (except verify_eddsa)
- Fully passed OPA tests: jwtbuiltins, jwtverifyhs[256|384|512], jwtverifyrsa
- Partially passed OPA tests: jwtdecodeverify (the ones that leverage io.jwt.verify_eddsa and io.jwt.encode_sign still do not pass)
- In order to enable customization of the crypto libraries behind JWT verification algorithms, this implementation introduces JWT Backend trait that describes common interface for JWT verification. The trait can be implemented using variable libraries, depending on user needs. The exact trait implementation is selected based on a feature provided to compiler. Currently only 1 inplementation of the JWT Backend is available (leverages OpenSSL as underlying crypto lib).

**To be implemented in the follow up PRs:**
- verify_eddsa support and token encoding builtins.

**Note** that I've also modified the implementation for opa tests in order to be able to run each test separately, not sure if it really belongs here.

Looking forward for a feedback

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -486,6 +486,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "foreign-types"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
+dependencies = [
+ "foreign-types-shared",
+]
+
+[[package]]
+name = "foreign-types-shared"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
+
+[[package]]
 name = "form_urlencoded"
 version = "1.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


wetpeanuts:
Hey @anakrish 

Thank you for the feedback

1. Guess I need to rebase to the latest main and use it. I'll give it a try.
2. I actually plan to introduce another Rust only JWT backend in the next iteration and use it as default. I agree that using OpenSSL by default is probably not the best idea since it might be missing in some environments.
3. I'll check how we can ensure using OpenSSL 3.x.

anakrish:
Hey @wetpeanuts,

Given MSFTs constraints on FIPS certification etc, we will likely want OpenSSL 3.x as the default. But having a rust only option is also good for other consumers of the crate.

Another approach to consider is whether a registration mechinasm can be provided to install an instance of the trait for use as the crypto implementation.

anakrish:
Also note that we are hardening regorus via strict clippy warnings. Do take a look at this: https://github.com/microsoft/regorus/pull/513#issuecomment-3702370679

wetpeanuts:
@anakrish Thank you for the hint. The data field you are talking is about regorus test coverage if I understand it correctly, is it correct? The problem for me lies in the original OPA tests from original OPA repo, that are fetched and being fed to the regorus engine in the opa test. With this limitation they do not pass. Since I assume changing the original test coverage is not an option I'm trying to find a way to overcome this on regorus side.

PR #499, opened 2025-11-07
feat: Detect incorrect multi-threaded use from c based ffi

anakrish:


bindings/ffi/src/engine.rs
```diff
@@ -318,9 +382,9 @@ pub extern "C" fn regorus_engine_set_gather_prints(
 #[no_mangle]
 pub extern "C" fn regorus_engine_take_prints(engine: *mut RegorusEngine) -> RegorusResult {
     let output = || -> Result<String> {
-        Ok(serde_json::to_string_pretty(
-            &to_ref(engine)?.engine.take_prints()?,
-        )?)
+        let engine = to_ref(engine)?;
+        let mut guard = engine.try_write()?;
```
dpokluda, line 386: >try_write [](http://example.com/codeflow?start=31&length=9)

Just curious - why is this write? Does it "clean" (the prints) afterwards? #Closed

anakrish: Rego has a `print` function whose behavior is to print to standard output. Kata Container's project uses these prints for debugging purposes, but they want the ability to capture these prints instead of displaying them in stdout.

So regorus Engine provides a way to toggle this behavior and to take any accumulated "prints". Taking the "prints" removes them from the Engine - the engine's gathered prints are cleared.

So yes, the write is needed since it is a mutating operation on the engine and yes it cleans the gathered prints.

PR #500, opened 2025-11-10
    feat!: Use num-bigint for large numbers

### LINKED ISSUE #165: Optimize Number representation
Currently, Regorus uses big floats to represent all numbers.
This needs to be optimized so that:
 - `u64`, `i64`, `f64` and big floats are used by default and big floats are used only as needed.
 - Big Float support needs to be under a feature
 - There needs to be a way to set the precision of Big Float operations.

anakrish:
feat!: Use num-bigint for large numbers
  
  - Supply chain: Use the popular num-bigint crate for handling large integers
  - Optimization: Handle f64, i64, u64 directly. These will be the most common instances of a number.
  
  OPA number semantics isn't clear.
  https://github.com/open-policy-agent/opa/issues/6281
  
  As part of this change, we update the following failing tests:
  - A local test that relies on what 15.3/3 evaluates to.
   With our current change, we round in a different direction than what OPA does, but consistent
   with Rust. We produce 5.1000000000000005 where as the OPA test expects 5.1.
   There is no clear definition in Rego of what the right answer is. Moreover, policies should not
   rely on exact floating point value comparison. Therefore this deviations is justified.
   The test is patched to pass.
  - Another local vm test that exercised 1.1 + 2.2
  - Another local vm test that exercises 5.5 - 2.2
  - An OPA test that expects that a large integer number say 10e308 is printed in exponent notation.
   num-bigint does not print using scientific notation and instead prints all the digits.
   The benefit of preserving this compatibility is not clear. We skip this test.
  - Doc tests that exercised handling floating point numbers with more than 15 (what f64 supports)
    digits of precision. There is no usecase for this scenario. The tests are updated to reflect
    the behavior.
      
  closes #165 
 Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>

### DIFF
--- a/src/number.rs
+++ b/src/number.rs
@@ -1,70 +1,158 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use alloc::str::FromStr;
-use core::cmp::{Ord, Ordering};
+use alloc::format;
+use alloc::string::{String, ToString};
+use core::cmp::Ordering;
 use core::fmt::{Debug, Formatter};
+use core::str::FromStr;
 
 use anyhow::{anyhow, bail, Result};
+use num_bigint::BigInt as NumBigInt;
+#[cfg(not(feature = "std"))]
+use num_traits::float::FloatCore;
+use num_traits::{One, Signed, ToPrimitive, Zero};
 
 use serde::ser::Serializer;
 use serde::Serialize;
 
 use crate::*;
 
-pub type BigInt = i128;
+pub type BigInt = NumBigInt;
 
-type BigFloat = scientific::Scientific;
-const PRECISION: scientific::Precision = scientific::Precision::Digits(100);
-
-#[derive(Clone, Debug, PartialEq)]
-pub struct BigDecimal {
-    d: BigFloat,
-}
-
-impl AsRef<BigFloat> for BigDecimal {
-    fn as_ref(&self) -> &BigFloat {
-        &self.d
-    }
-}
-
-impl AsMut<BigFloat> for BigDecimal {
-    fn as_mut(&mut self) -> &mut BigFloat {
-        &mut self.d
-    }
-}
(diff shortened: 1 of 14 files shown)


src/number.rs
```diff
@@ -1,239 +1,460 @@
...
+
+        if value.is_negative() {
+            if let Some(i) = value.to_i64() {
+                return Number::Int(i);
+            }
+        } else if let Some(u) = value.to_u64() {
+            return Number::UInt(u);
+        } else if let Some(i) = value.to_i64() {
+            return Number::Int(i);
+        }
```
dekomissMSFT, line 47: nit: not sure if it matters, but should we try converting to i64 first and then u64? otherwise won't the i64 clause never be hit?

src/number.rs
```diff
@@ -254,250 +512,464 @@ impl Number {
...
-                Ok(Big(BigDecimal::from(c).into()))
+        if rhs.is_zero() {
+            bail!("division by zero");
+        }
+
+        if matches!(self, Number::Float(_)) || matches!(rhs, Number::Float(_)) {
+            return Ok(Number::Float(self.to_f64_lossy() / rhs.to_f64_lossy()));
+        }
+
+        match (&self, rhs) {
```
dekomissMSFT, line 570: > match (&self, rhs) [](http://example.com/codeflow?start=7&length=19)

seeing a lot of new code here - are there nuances to all of these combinations that we should add tests for? or is it all mostly boilerplate and a result of the different combinations of number types we're supporting?

src/number.rs
anakrish: Good observation. Let me reorder.

src/number.rs
anakrish: There are some nuances as well as boilerplate. Let me add the tests in the next PR.

PR #506, opened 2025-11-24
feat!: Rego -> RVM Compiler and extensive testsuite

anakrish:


src/languages/rego/compiler.rs
anakrish, line 1: This entire file can be ignored. This monolithic compiler implementation is split up into separate files in subsequent checkins and finally removed.

PR #515, opened 2025-12-02
feat: Handle more OPA semantics in RVM compiler

anakrish:
1. out parameter syntax.
2. not expression with out parameters

Also add span information to compiler errors.


src/languages/rego/compiler/function_calls.rs
```diff
@@ -16,61 +29,160 @@ impl<'a> Compiler<'a> {
...
+            Some(2)
+        } else {
+            builtins::BUILTINS
+                .get(name)
+                .map(|(_, arity)| *arity as usize)
+        }
+    }
+}
+
+impl<'a> Compiler<'a> {
```
dpokluda, line 150: Just curious - why do we end the `impl` scope and then define it atain?

dpokluda (approved):
:shipit:

src/languages/rego/compiler/function_calls.rs
anakrish: In this case, no special reason other than me forgetting to remove the extra block after an attempted (and aborted) refactor.
I will do a final cleanup later since if I make any change now, you will have to reapprove.

PR #516, opened 2025-12-03
Handle more OPA semantics in RVM and compiler

anakrish:


src/languages/rego/compiler/references.rs
```diff
@@ -24,26 +24,37 @@ pub(super) enum AccessComponent {
...
     /// Chain of field accesses - either literal field names or dynamic expressions
     pub(super) components: Vec<AccessComponent>,
 }
 
 impl ReferenceChain {
     /// Get the static prefix path (all literal components from the start)
-    pub(super) fn get_static_prefix(&self) -> Vec<&str> {
-        let mut prefix = vec![self.root.as_str()];
+    pub(super) fn get_static_prefix(&self) -> Option<Vec<&str>> {
+        let ReferenceRoot::Variable(root) = &self.root else {
```
dpokluda, line 46: >let ReferenceRoot::Variable(root) = &self.root else [](http://example.com/codeflow?start=8&length=51)

Good, but I need to get used to this syntax (I am used to C# pattern matching `if (a is Pattern) ...` but in Rust it's opposite `let Pattern = a`! #Resolved

PR #524, opened 2025-12-13
fix: Integer underflow in lexer error message formatting

### LINKED ISSUE #518: [potential bug] panics when there is a new line before dot
**regorus version: 0.5.0**

[Test1] normal rego input:

```
use regorus::Engine;

fn main() {
    // Create a new Rego engine
    let mut engine = Engine::new();

    let policy = r#"
package example

# Policy using regex.match
result = output if {
    text := "hello world"
    output := regex.match("^hello", text)
}
"#;

    // Add the policy to the engine
    match engine.add_policy("example.rego".to_string(), policy.to_string()) {
        Ok(_) => println!("Policy added successfully"),
        Err(e) => println!("Error adding policy: {}", e),
    }

    // Set input data (optional for this example)
    match engine.set_input_json("{}") {
        Ok(_) => println!("Input set successfully"),
        Err(e) => println!("Error setting input: {}", e),
    }

    // Evaluate the policy
    match engine.eval_query("data.example.result".to_string(), false) {
        Ok(result) => println!("Result: {:#?}", result),
        Err(e) => println!("Error evaluating query: {}", e),
    }
}
```

Result: Ok, as expected

[Test2] replace the rego with (new line after regex):

```
result = output if {
    text := "hello world"
output := regex
.match("^hello", text)
}
"#;
```
Result: Error, as expected

[Test3] replace the rego with (new line after regex.):
```
result = output if {
    text := "hello world"
output := regex.
match("^hello", text)
}
"#;
```
Result: Panic.

Image

Shouldn't the third test also return an error as the second test?

hkrutzer:
See #518 

The example now produces:

```
Error adding policy:
--> example.rego:8:0
  |
8 | match("^hello", text)
  | ^
error: invalid whitespace between . and identifier
Input set successfully
Result: QueryResults {
    result: [],
}
```

fixes #518

### DIFF
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -203,7 +203,7 @@ impl Source {
 
         let line_str = format!("{line}");
         let line_num_width = line_str.len() + 1;
-        let col_spaces = col as usize - 1;
+        let col_spaces = (col as usize).saturating_sub(1);
 
         format!(
             "\n--> {}:{}:{}\n{:<line_num_width$}|\n\
--- a/tests/interpreter/cases/call/basic.yaml
+++ b/tests/interpreter/cases/call/basic.yaml
@@ -75,4 +75,13 @@ cases:
         }
     query: data.test
     error: expects numeric argument.
-    
+
+  - note: multiline-member-access-should-error
+    data: {}
+    modules:
+      - |
+        package test
+        output := regex.
+        match("^a", "b")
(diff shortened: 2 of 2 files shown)


anakrish (approved):
LGTM. Thanks for the fix, @hkrutzer.

PR #526, opened 2025-12-16
Quick fixes

anakrish:


scripts/pre-push
```diff
@@ -10,31 +10,31 @@ if [ -f Cargo.toml ]; then
...
    # Ensure that the public API works.
-   cargo test -r --doc
+   cargo test --doc
 
    # Ensure that no_std build succeeds.
    # Build for a target that has no std available.
    if command -v rustup > /dev/null; then
        rustup target add thumbv7m-none-eabi
-       (cd tests/ensure_no_std; cargo build -r --target thumbv7m-none-eabi --no-default-features --features opa-no-std)
+       (cd tests/ensure_no_std; cargo build -r --target thumbv7m-none-eabi)
```
tjons, line 19: I tested this change and it does fix the issues I was seeing locally with this target. Previously, cargo was complaining about the `opa-no-std` feature.

PR #527, opened 2025-12-19
Checked indexing

anakrish:


src/interpreter.rs
```diff
@@ -359,16 +359,20 @@ impl Interpreter {
...
 
     // Helper methods for working with ExprLookup
-    fn set_loop_var_value(&mut self, expr: &ExprRef, value: Value) {
+    fn set_loop_var_value(&mut self, expr: &ExprRef, value: Value) -> Result<()> {
         let module_idx = self.current_module_index;
         let expr_idx = expr.eidx();
-        self.loop_var_values.set(module_idx, expr_idx, value);
+        self.loop_var_values
+            .set_checked(module_idx, expr_idx, value)
+            .map_err(|err| anyhow!("internal error: loop var indices out of bounds: {err}"))
```
dpokluda, line 367: >.map_err(|err| anyhow!("internal error: loop var indices out of bounds: {err}")) [](http://example.com/codeflow?start=12&length=80)

I'm still new to this. My understanding is that you want to replace the original error with your own error (with the text that says what's inside the brackets). By doing that might loose the context about the original error that actually caused the issue (except the fact that we include it within the error message). 

My understanding:
- `map_err`: to change one error into another error (loosing the original error context)
- `context`: add human-friendly context while preserving the original error as the source

anakrish: You are absolutely right about anyhow's with_context being better about preserving and presenting contexts nicely.

Note that `set_checked` returns core::Result which is different from `anyhow::Result` and does not have `with_context`.

We plan to move away from anyhow (https://github.com/microsoft/regorus/issues/161) since it is not intended to be used by libraries. In newer code, whenever possible I've been using custom error types/result. When we address the above issue, you are right, it would be nice to implement our own with_context like functionality that will preserve contexts.

anakrish:
Thanks for the proactive review @dpokluda !

PR #530, opened 2025-12-23
chore: Fix lint errors in lookup.rs

anakrish:
- Made the lookup module crate-visible to address clippy’s redundant visibility lint.
- Replaced unchecked as casts with a fallible usize_from_u32 helper and propagate conversion errors in lookup accessors.
- Switched LookupIndexError to implement core::error::Error for no_std correctness.
- Fixed the pattern type mismatch by matching on the value in the Display impl.
- Promoted trivial helpers to const fn (new, module_len) per clippy suggestions.
- Centralized bounds-checked slot access via slot_ref/slot_mut to keep getters/clearers lint-clean and avoid unchecked indexing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,7 +135,7 @@ pub mod languages {
 }
 
 mod lexer;
-mod lookup;
+pub(crate) mod lookup;
 mod number;
 mod parser;
 mod policy_info;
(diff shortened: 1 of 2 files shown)


dpokluda (approved):
:shipit:

PR #531, opened 2025-12-26
chore: Harden lexer bounds and span handling

anakrish:
- Document arithmetic safety assumptions and add explicit lexer limits for columns, file size (1 MiB), and line count. Realistic policies will be well within these bounds.
- Use checked arithmetic to prevent overflow underflow.
- Avoid var name shadowing.
- Misc clippy lints


dpokluda (approved):
:shipit:

PR #539, opened 2026-01-08
feat!: Ability to limit policy evaluation time

anakrish:
- Introduce ExecutionTimer/ExecutionTimerConfig to allow limiting evaluating time.
  - To amortize time checking costs, checking interval can be configured via the notion of work units
  - A global fallback time limit can be set to universally limit all evaluation in addition to engine level limit setting.
  - Implement limnits in interpreter and RVM. In RVM, also handle suspend/resume so that time during pause is not counted.
  - Add engine-level APIs to set/clear per-engine timer configuration and apply global fallback defaults.
  - Surface execution-time limits through FFI and C# bindings
  - Add C# tests and example usage to validate engine overrides, global fallback behavior, and compiled policy enforcement.
  - Expand docs for execution-time limit
  - Add interpreter YAML cases and VM unit tests for time-limit behavior and deterministic time sources.

### DIFF
--- a/bindings/csharp/Regorus/CompiledPolicy.cs
+++ b/bindings/csharp/Regorus/CompiledPolicy.cs
@@ -203,5 +203,14 @@ namespace Regorus
                 }
             }
         }
+
+        private void UseHandle(Action<IntPtr> action)
+        {
+            UseHandle<object?>(handlePtr =>
+            {
+                action(handlePtr);
+                return null;
+            });
+        }
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ bindings/*/target
 # C# build folders
 **bin
 **obj
+bindings/csharp/.nuget/
 
 # Bundler binstubs regenerated during ruby setup
 bindings/ruby/bin/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1269,6 +1269,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "spin",
  "test-generator",
  "thiserror",
  "url",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,6 +105,7 @@ thiserror = { version = "2.0", default-features = false }
 data-encoding = { version = "2.8.0", optional = true, default-features=false, features = ["alloc"] }
 num-bigint = { version = "0.4", default-features = false }
 num-traits = { version = "0.2", default-features = false }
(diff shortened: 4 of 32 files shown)


bindings/csharp/Regorus/CompiledPolicy.cs
```diff
@@ -203,5 +203,14 @@ private T UseHandle<T>(Func<IntPtr, T> func)
...
         }
+
+        private void UseHandle(Action<IntPtr> action)
+        {
+            UseHandle<object?>(handlePtr =>
+            {
+                action(handlePtr);
+                return null;
+            });
+        }
```
dpokluda, line 214: nit: I don't know how hot is this code path but we are boxing `null` as `object?`. We could instead copy/paste the UseHandle method without the return:

``` csharp
private void UseHandle(Action<IntPtr> action)
{
    var handle = GetHandleForUse();
    bool addedRef = false;
    try
    {
        handle.DangerousAddRef(ref addedRef);
        var pointer = handle.DangerousGetHandle();
        if (pointer == IntPtr.Zero)
        {
            throw new ObjectDisposedException(nameof(CompiledPolicy));
        }

        action(pointer);
    }
    finally
    {
        if (addedRef)
        {
            handle.DangerousRelease();
        }
    }
}
```

dpokluda: An alternative option would be to refactor the handle handling :-) logic to a separate method and call it from both:
``` csharp
private T UseHandle<T>(Func<IntPtr, T> func)
{
    T? result = default!;
    WithHandle(ptr => result = func(ptr));
    return result!;
}

private void UseHandle(Action<IntPtr> action)
{
    WithHandle(action);
}

private void WithHandle(Action<IntPtr> body)
{
    var handle = GetHandleForUse();
    bool addedRef = false;
    try
    {
        handle.DangerousAddRef(ref addedRef);
        var pointer = handle.DangerousGetHandle();
        if (pointer == IntPtr.Zero)
        {
            throw new ObjectDisposedException(nameof(CompiledPolicy));
        }

        body(pointer);
    }
    finally
    {
        if (addedRef)
        {
            handle.DangerousRelease();
        }
    }
}
```

anakrish: Created https://github.com/microsoft/regorus/issues/554 to track this. There is one C# specific PR remaining (expose VM to C#) that will wrap up the recent stream of work to improve robustness at Azure scale (super strict clippy lints, memory limits, executions limits). I will address this along with that.

PR #563, opened 2026-01-29
chore(release): publish vendored mimalloc crates

anakrish:
- Prefix regorus- to mimalloc crates and add MIT licenses
- alias dependencies to avoid code changes
- add versions and release-plz publish entries
- update Cargo.lock files for new crate names


dpokluda (approved):
:shipit:

PR #564, opened 2026-01-30
chore: Gate exports using allocator-limits feature

anakrish:
This is needed to publish regorus-mimalloc

### DIFF
--- a/mimalloc/src/mimalloc.rs
+++ b/mimalloc/src/mimalloc.rs
@@ -29,9 +29,9 @@ unsafe impl GlobalAlloc for Mimalloc {
     }
 
     #[inline]
-    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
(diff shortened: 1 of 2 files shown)


mimalloc/src/mimalloc.rs
```diff
@@ -29,9 +29,9 @@ unsafe impl GlobalAlloc for Mimalloc {
     }
 
     #[inline]
-    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
```
dpokluda, line 32: >_ [](http://example.com/codeflow?start=43&length=1)

Just curious, why the underscore prefix?

anakrish: The layout parameter is used only conditionally when the limits feature is enabled. So clippy raises an error about unused parameter when the code-path is not enabled. The way to fix it is to prefix with _. 

I debated a bit with copilot on whether there are other ways to fix, but finally this way seems to be the most idiomatic.

PR #577, opened 2026-02-09
feat(rbac)!: add Azure RBAC engine, FFI API, and cross-language tests

anakrish:
- add Azure RBAC condition interpreter and builtin evaluation in core (expressions, parser updates, evaluator, and test harness)
- introduce comprehensive RBAC YAML test suites and coverage for i
  - action/suboperation
  - strings
  - numbers
  - bools
  - IP
  - GUID
  - dates
  - times
  - lists
  - quantifiers (ForAnyOfAnyValues, ForAllOfAllValues)
- expose RBAC evaluation through FFI with an `rbac` feature flag enabled by default
- add C# `RbacEngine` wrapper + P/Invoke entrypoint and document usage in C# README
- expand C# tests to execute all RBAC YAML cases with per-case logging
- wire test assets into C# test output and centralize YAML dependency versions


src/languages/azure_rbac/builtins/actions.rs
```diff
@@ -0,0 +1,41 @@
...
+// - action:  "Microsoft.Storage/storageAccounts/read"
+//   pattern: "Microsoft.Storage/storageAccounts"     => false (segment count mismatch)
+pub(super) fn action_matches_pattern(action: &str, pattern: &str) -> bool {
+    // Split action and pattern into fixed segments (delimited by '/').
+    let action_parts: alloc::vec::Vec<&str> = action.split('/').collect();
+    let pattern_parts: alloc::vec::Vec<&str> = pattern.split('/').collect();
+
+    // Require the same number of segments for a match.
+    if action_parts.len() != pattern_parts.len() {
+        return false;
```
dekomissMSFT, line 24: I don't think wildcard matching is limited to individual segments in RBAC - for example, a pattern of Microsoft.Storage/* would allow Microsoft.Storage/storageAccounts/read, but this method would return false for it

I think wildcards also apply to nested resources, e.g. Microsoft.Storage/*/read would match Microsoft.Storage/storageAccounts/read and Microsoft.Storage/storageAccounts/blobServices/read

bindings/csharp/README.md
```diff
@@ -104,3 +104,49 @@ vm.SetInputJson(Input);
...
+    "private_endpoint": null,
+    "subnet": null,
+    "utc_now": "2023-05-01T12:00:00Z"
+  },
+  "action": "Microsoft.Storage/storageAccounts/read",
+  "suboperation": null
+}
+""";
+
+var allowed = RbacEngine.EvaluateCondition(Condition, ContextJson);
```
dekomissMSFT, line 150: >EvaluateCondition [](http://example.com/codeflow?start=25&length=17)

how would you pass the actions/notactions here?

dekomissMSFT (approved):
:shipit:

bindings/csharp/Directory.Packages.props
dekomissMSFT, line 4: should we bump the versions or do you prefer to do it in a separate PR?

src/languages/azure_rbac/builtins/actions.rs
anakrish: I will address this in another PR as we test extensively.

bindings/csharp/README.md
anakrish: This PR is mainly focused on the condition expression sub language. NoActions, assignments etc will be part of a higher-level rbac evaluator.

bindings/csharp/Directory.Packages.props
anakrish: I can bump it up when we make the next version with more changes.

PR #664, opened 2026-04-07
fix: RVM evaluation of default-only rules

kusha:
### Problem

Rego rules that have only a `default` declaration with no conditional body (e.g., `default deny := {"result": false}`) return `Undefined` in the RVM instead of the default value. The interpreter handles this correctly — the bug is RVM-only.

### Root cause

Two gaps in the RVM compiler and VM:

1. **Compiler (`compute_rule_type`)**: When a rule path exists only in the `default_rules` map (no entries in `rules`), the compiler errors with "no definitions found for rule path" instead of recognizing it as a `Complete` rule.

2. **Compiler (`compile_worklist_rule`)**: The `else` branch (no definitions to compile) was a no-op — it didn't allocate register slots or add the rule to the data tree, so `finish()` would panic or the rule was undiscoverable at eval time.

3. **VM (`execute_call_rule_common` / `execute_call_rule_suspendable`)**: When `rule_definitions` is empty, both functions immediately returned `Undefined` without checking the `default_literal_index` for a pre-compiled default value.

### Fix

- `compute_rule_type`: return `Complete` when the rule path exists in `default_rules` but not `rules`
- `compile_worklist_rule` else branch: size `rule_num_registers`, add the rule to the program's data tree
- `execute_call_rule_common` + `execute_call_rule_suspendable`: look up `default_literal_index` and return the stored literal before falling back to `Undefined`

### Tests

- 3 new RVM test cases in `default_rules.yaml`: `default_only_rule_bool`, `default_only_rule_object`, `default_only_rule_with_entry_point`
- 3 matching interpreter test cases in `default/basic.yaml` (confirming the interpreter already handles these correctly)

### DIFF
--- a/src/languages/rego/compiler/rules.rs
+++ b/src/languages/rego/compiler/rules.rs
@@ -38,6 +38,11 @@ impl<'a> Compiler<'a> {
 
     pub(super) fn compute_rule_type(&self, rule_path: &str) -> Result<RuleType> {
         let Some(definitions) = self.policy.inner.rules.get(rule_path) else {
+            // Default-only rules (e.g., `default deny := true`) have no regular definitions
+            // in the `rules` map — they only exist in `default_rules`. Treat them as Complete.
+            if self.policy.inner.default_rules.contains_key(rule_path) {
+                return Ok(RuleType::Complete);
+            }
             return Err(CompilerError::General {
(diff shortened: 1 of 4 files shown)


anakrish (approved):
Thanks for the fix @kusha !

PR #667, opened 2026-04-08
feat(compiler): support registered host-await builtins for natural function call syntax

kusha:
## Summary

Extends the RVM compiler to accept a list of host function names at compile time. Calls to registered names are compiled directly into `HostAwait` instructions, allowing policy authors to write `lookup(input.account_id)` instead of `__builtin_host_await(input.account_id, "lookup")`.

## Motivation

1. **No compile-time validation of host-supported identifiers**: The identifier passed to `__builtin_host_await` is a runtime string. Typos like `"lokup"` instead of `"lookup"` are silently accepted by the compiler and only discovered at runtime when the host rejects the identifier.

2. **Wrapper workaround breaks caller identification**: Defining a wrapper `lookup(x) := __builtin_host_await(x, "lookup")` compiles the `HostAwait` instruction inside the wrapper module, not the calling module — breaking caller identification for authorization and auditing.

3. **Host can override standard builtins**: Registering a name that matches a standard Rego builtin (e.g. `time.parse_duration_ns`) lets the host intercept that call and provide its own implementation.

### Compiler (`src/languages/rego/compiler/`)

- `register_host_await_builtin()` on `Compiler` to declare host function names. Validates `arg_count == 1` and rejects the reserved name `__builtin_host_await`.
- `compile_from_policy_with_host_await()` entry point that accepts the builtin list. Existing `compile_from_policy()` delegates with an empty list.
- `determine_call_target()` extended with resolution order: explicit `__builtin_host_await` → registered host-await → user-defined → standard builtin.
- Both explicit and registered paths emit identical `HostAwait { dest, arg, id }` bytecode.

### Documentation (`docs/rvm/`)

- **`instruction-set.md`**: Registered builtin syntax, resolution order, argument handling.
- **`architecture.md`**: Updated to describe both emission paths.

### Tests (`tests/rvm/rego/`)

- Extended test harness with `HostAwaitBuiltinSpec` and argument assertion.
- 9 YAML test cases: suspend/resume, run-to-completion, multiple names, queue, shadowing, object packing, arg_count rejection, reserved name rejection, standard builtin override.

## Constraints

- **`arg_count` must be 1**: The `HostAwait` instruction carries a single argument register. Use object packing to pass multiple values: `lookup({"key1": v1, "key2": v2})`. This can be lifted in the future by having the compiler auto-pack multiple arguments into an array or object.
- **`__builtin_host_await` is reserved**: Attempting to register this name produces a compile-time error. It is handled by a dedicated code path (explicit 2-argument form) and cannot be overridden.

## Limitations

- **No conflict detection**: A registered name that collides with a standard builtin silently wins. No warning is emitted.
- **`__builtin_host_await` still exists**: The raw builtin remains functional. Policy authors who discover it can use arbitrary identifier strings. However, the host controls how identifiers are resolved and can reject or error on unexpected identifiers at runtime.

### DIFF
--- a/src/languages/rego/compiler/function_calls.rs
+++ b/src/languages/rego/compiler/function_calls.rs
@@ -17,6 +17,7 @@ use crate::lexer::Span;
 use crate::rvm::instructions::{BuiltinCallParams, FunctionCallParams};
 use crate::rvm::Instruction;
 use crate::utils::get_path_string;
+use crate::value::Value;
 use alloc::{format, string::ToString, vec::Vec};
 
 enum CallTarget {
@@ -127,21 +128,49 @@ impl<'a> Compiler<'a> {
                 self.emit_instruction(Instruction::BuiltinCall { params_index }, &span);
             }
             CallTarget::HostAwait { .. } => {
-                if arg_regs.len() != 2 {
-                    return Err(CompilerError::General {
-                        message: format!(
-                            "__builtin_host_await expects 2 arguments, got {}",
-                            arg_regs.len()
-                        ),
+                let (arg_reg, id_reg) = if original_fcn_path == "__builtin_host_await" {
+                    // Explicit __builtin_host_await(arg, id) — 2 arguments
+                    if arg_regs.len() != 2 {
+                        return Err(CompilerError::General {
+                            message: format!(
+                                "__builtin_host_await expects 2 arguments, got {}",
+                                arg_regs.len()
+                            ),
+                        }
+                        .at(&span));
                     }
-                    .at(&span));
-                }
+                    (arg_regs[0], arg_regs[1])
+                } else {
+                    // Registered host-awaitable builtin — identifier is the function name
+                    if arg_regs.is_empty() {
+                        return Err(CompilerError::General {
+                            message: format!(
+                                "host-awaitable builtin '{}' expects at least 1 argument, got 0",
+                                original_fcn_path
+                            ),
+                        }
+                        .at(&span));
+                    }
+                    let id_reg = self.alloc_register();
+                    let literal_idx = self.add_literal(Value::String(original_fcn_path.into()));
+                    self.emit_instruction(
+                        Instruction::Load {
+                            dest: id_reg,
+                            literal_idx,
+                        },
+                        &span,
+                    );
+                    // HostAwait carries a single arg register; registered builtins
+                    // are restricted to arg_count == 1 at registration time, so
+                    // arg_regs[0] is the only argument.
+                    (arg_regs[0], id_reg)
+                };
 
                 self.emit_instruction(
                     Instruction::HostAwait {
                         dest,
-                        arg: arg_regs[0],
-                        id: arg_regs[1],
+                        arg: arg_reg,
+                        id: id_reg,
                     },
                     &span,
                 );
@@ -189,6 +218,13 @@ impl<'a> Compiler<'a> {
         if original_fcn_path == "__builtin_host_await" {
             return Ok(CallTarget::HostAwait {
                 expected_args: Some(2),
+            });
+        }
+
+        // Check registered host-awaitable builtins
+        if let Some(&arg_count) = self.host_await_builtins.get(original_fcn_path) {
+            return Ok(CallTarget::HostAwait {
+                expected_args: Some(arg_count),
             });
         }
 
--- a/src/languages/rego/compiler/function_calls.rs
+++ b/src/languages/rego/compiler/function_calls.rs
@@ -142,11 +142,12 @@ impl<'a> Compiler<'a> {
                     (arg_regs[0], arg_regs[1])
                 } else {
                     // Registered host-awaitable builtin — identifier is the function name
-                    if arg_regs.is_empty() {
+                    if arg_regs.len() != 1 {
                         return Err(CompilerError::General {
                             message: format!(
-                                "host-awaitable builtin '{}' expects at least 1 argument, got 0",
-                                original_fcn_path
+                                "host-awaitable builtin '{}' expects exactly 1 argument, got {}",
+                                original_fcn_path,
+                                arg_regs.len()
                             ),
                         }
                         .at(&span));
--- a/src/languages/rego/compiler/function_calls.rs
+++ b/src/languages/rego/compiler/function_calls.rs
@@ -246,6 +246,16 @@ impl<'a> Compiler<'a> {
         // restricted to arg_count == 1 at registration time (see
         // `Compiler::register_host_await_builtin`), so the variant doesn't
         // need to carry an arity — it's fixed at 1.
+        //
+        // We deliberately match against `original_fcn_path` only, not
+        // `full_fcn_path`. Registration intercepts the *unqualified* call
+        // form (e.g. `lookup(x)` inside the policy's own package). A
+        // package-qualified call like `data.other.lookup(x)` is left to
+        // resolve through the normal user-defined / builtin path, so a
+        // registered name does not leak into unrelated packages that
+        // happen to expose a rule with the same identifier. This is
+        // documented on `register_host_await_builtin`; the
+        // `registered_host_await.yaml` suite pins the behavior.
         if self.host_await_builtins.contains_key(original_fcn_path) {
             return Ok(CallTarget::RegisteredHostAwait {
                 identifier: original_fcn_path.to_string(),
--- a/src/languages/rego/compiler/function_calls.rs
+++ b/src/languages/rego/compiler/function_calls.rs
@@ -18,8 +18,19 @@ use crate::rvm::instructions::{BuiltinCallParams, FunctionCallParams};
 use crate::rvm::Instruction;
 use crate::utils::get_path_string;
 use crate::value::Value;
-use alloc::{format, string::ToString, vec::Vec};
+use alloc::{
+    format,
+    string::{String, ToString},
+    vec::Vec,
+};
 
+/// Resolved destination of a Rego function-call expression. Produced by
+/// [`Compiler::determine_call_target`] and consumed by
+/// [`Compiler::compile_function_call`] to choose which instruction to emit.
+/// Carrying the discrimination in the type (rather than re-matching on a
+/// magic name at the emit site) keeps the host-await handling honest under
+/// future refactors — the compiler will refuse to build if a new variant is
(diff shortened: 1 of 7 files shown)


tests/rvm/rego/cases/registered_host_await.yaml
```diff
@@ -0,0 +1,240 @@
+cases:
```
anakrish, line 1: More test cases to lock down behavior:

- Out param syntax: e.g: `lookup(in, out)`
- Test case that uses __builtin_host_await and registerd builtins in same policy
- Empty host await builtins

src/languages/rego/compiler/rules.rs
```diff
@@ -186,8 +181,20 @@ impl<'a> Compiler<'a> {
...
         entry_points: &[&str],
+    ) -> Result<Arc<Program>> {
+        Self::compile_from_policy_with_host_await(policy, entry_points, &[])
+    }
+
+    /// Compile from a CompiledPolicy to RVM Program with registered host-awaitable builtins.
+    pub fn compile_from_policy_with_host_await(
+        policy: &CompiledPolicy,
+        entry_points: &[&str],
+        host_await_builtins: &[(&str, usize)],
```
anakrish, line 192: Scenarios to define behavior for:
- Duplicate entries in host_await_builtins
- Empty host_await_builtins
- A name is empty or whitespace

src/languages/rego/compiler/function_calls.rs
```diff
@@ -127,21 +128,50 @@ impl<'a> Compiler<'a> {
...
+                            ),
+                        }
+                        .at(&span));
                     }
-                    .at(&span));
-                }
+                    (arg_regs[0], arg_regs[1])
+                } else {
+                    // Registered host-awaitable builtin — identifier is the function name
+                    if arg_regs.len() != 1 {
```
anakrish, line 145: Need to define behavior when registered host await builtin shadows a user write policy rule (both regular and function)

src/languages/rego/compiler/function_calls.rs
```diff
@@ -192,6 +222,13 @@ impl<'a> Compiler<'a> {
             });
         }
 
+        // Check registered host-awaitable builtins
+        if let Some(&arg_count) = self.host_await_builtins.get(original_fcn_path) {
```
anakrish, line 226: We need to detect shadowing here.

src/languages/rego/compiler/function_calls.rs
```diff
@@ -192,6 +222,13 @@ impl<'a> Compiler<'a> {
...
+            return Ok(CallTarget::HostAwait {
```
anakrish, line 227: Maybe it might be clear to distinguish between the builtin and the custom registered builtins via a new variant say RegisteredHostAwait. That way the rest of the code doesn't need to depend on the builtin's name (__builtin_host_await) to distinguish between builtin and registered.

anakrish (commented):
Overall looks good.

tests/rvm/rego/mod.rs
```diff
@@ -279,6 +310,14 @@ fn compile_and_run_rvm_with_all_entry_points(
                                         identifier
                                     )
                                 })?;
+                            if let Some(expected) = expected_args {
+                                let actual = process_value(argument)?;
+                                assert_eq!(
+                                    actual, expected,
+                                    "HostAwait argument mismatch for {:?}: expected {:?}, got {:?}",
+                                    identifier, expected, actual
+                                );
```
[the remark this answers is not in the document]
kusha: Fixed

tests/rvm/rego/cases/registered_host_await.yaml
```diff
@@ -0,0 +1,240 @@
...
+      - name: kv_store
+        arg_count: 2
+    modules:
+      - |
+        package demo
+        import rego.v1
+
+        result := kv_store(input.key, input.value)
+    query: data.demo.result
+    # Registration panics because arg_count must be 1.
```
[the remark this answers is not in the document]
kusha: Fixed

tests/rvm/rego/cases/registered_host_await.yaml
kusha: Fixed: out works, test cases extended, empty

src/languages/rego/compiler/rules.rs
kusha: Fixed: empty means no registered

src/languages/rego/compiler/function_calls.rs
kusha: Can we move it out of scope for this PR? Shadowing already happens for `__builtin_host_await`. Also we need to decide on approach. IMO compilation check is not great choice as it makes compiled policies dependent on engine state. The easiest would be to prioritize user defined if it is present.

src/languages/rego/compiler/function_calls.rs
kusha: See the comment above

src/languages/rego/compiler/function_calls.rs
kusha: Fixed, but please do a thorough review of 357fbf309f2719ed22c0f0154cff4537b34a4aa7

anakrish:
@copilot Review this PR using all the skills in this repo. Use a separate agent for each skill.

anakrish:
Deep review found a few issues worth addressing before merge. I reviewed this with separate passes for value-flow, safety/API, type conversion, test adequacy, adversarial verification, and an independent deep-review agent.

### Findings

1. **Medium: fully-qualified calls can bypass registered host-await shadowing**

   The docs/tests say registered names shadow user-defined functions and builtins, but `determine_call_target` only checks `host_await_builtins` using `original_fcn_path`:

   ```rust
   if self.host_await_builtins.contains_key(original_fcn_path) {
       return Ok(CallTarget::RegisteredHostAwait {
           identifier: original_fcn_path.to_string(),
       });
   }
   ```

   For `package demo`, registering `("resolve", 1)` intercepts `resolve("k")`, but a policy can call `data.demo.resolve("k")`. In that case `original_fcn_path` is `data.demo.resolve`, so it does not match the registered key `resolve`; the call then falls through to the user-defined function path. Either the compiler should also account for the package-qualified form, or the docs/tests should explicitly state that registered host-await builtins only intercept unqualified calls.

2. **Low: suspendable HostAwait arg checks reprocess runtime values as YAML fixtures**

   In `tests/rvm/rego/mod.rs`, the suspendable harness does:

   ```rust
   let actual = process_value(argument)?;
   if actual != expected {
       return Err(anyhow::anyhow!(...));
   }
   ```

   `argument` is already a runtime VM `Value`; `process_value` is intended for YAML fixture decoding. This can turn legitimate runtime values into fixture sentinels before comparison. Example: policy passes `lookup("#undefined")`; YAML expected `args: "#undefined"` becomes `Value::Undefined`, and the runtime string is also converted to `Value::Undefined`, so the test can pass for the wrong reason. Compare `argument.clone()` directly to the already-processed expected value instead.

3. **Low: run-to-completion `args:` are silently ignored**

   `build_host_await_response_vec` parses expected args via the shared map builder and then drops them:

   ```rust
   .map(|(id, values)| (id, values.into_iter().map(|(_, output)| output).collect()))
   ```

   A run-to-completion YAML case with `host_await_responses: [{ id: "translate", args: "WRONG", value: "hola" }]` would still pass if the output matches, because the argument is never checked. Consider rejecting `args` for run-to-completion mode, documenting the limitation clearly, or adding instrumentation so RTC tests can validate payloads too.

4. **Low: HostAwait tests do not fail on unused expected responses**

   The suspendable harness pops responses when host-await calls occur, but does not assert that all response queues are empty after completion. A test can provide two expected responses for `lookup` while the policy calls it once; the leftover response is ignored. This can hide missing calls or too-few HostAwait executions. After completion, assert all suspendable response queues were consumed; add an equivalent check or explicit limitation for RTC mode.

5. **Low: whitespace-padded registered names create unreachable registrations**

   `register_host_await_builtin` rejects all-whitespace names via `name.trim().is_empty()`, but accepts names like `" lookup"` or `"lookup "`. Those get inserted into `host_await_builtins`, but normal Rego function-call paths will produce `lookup`, so the registration is effectively dead. Consider rejecting `name != name.trim()` or normalizing before insertion.

I dropped the general “registered names shadow user/standard builtins” concern because the current PR explicitly documents and tests that precedence as intended behavior. The FQN case above is narrower: the documented shadowing behavior does not appear to hold for fully-qualified call syntax.

anakrish:
@zicklag Does this PR address your needs in regards to wrapping the host await builtin?

zicklag:
Ooh, yeah, from the PR description this looks really useful. Especially being able to move some of the checks to compile time checks instead of runtime checks.

It is worth mentioning that _maybe_ I'd sometimes want to add wrapper functions in the rego policy just to allow you to call functions with more than one arg, ( and just forward that to the single arg async builtin ) but that doesn't reduce the value of this, and passing in a single object as an arg might be best in many cases anyway.

kusha:
@anakrish all addressed.  Per-finding:

1. (FQN bypass): Documented as intended, not changed. Matching the bare name is consistent with how the explicit form and  standard builtins already resolve (only user-defined functions  use the package-qualified path). Expanding registration to  qualified paths would let a registered name leak into any  package exposing a same-named rule. Clarified   register_host_await_builtin 's doc +  instruction-set.md ,  and added tests pinning.

2. (arg double-processing): Fixed — the harness now compares  the runtime  argument  directly against the already-decoded  expected value instead of re-running  process_value . Added a  set-valued arg regression test that fails under the old code  ( unexpected set... ) and passes now. 

3. (RTC  args:  ignored):  args:  is now rejected for  run-to-completion fixtures at load time (validation only  exists in suspendable mode); also stopped building the RTC vec  for suspendable cases. Added a pinning case.

4. (unconsumed responses): Not fixed, I propose to defer to a separate PR — it  predates this work (the suspendable harness from asserted drain). I can file an issue and follow up if needed.

5. (whitespace-padded names): Now reject  name !=  name.trim() , not just all-whitespace, with leading/trailing  test cases.

anakrish:
@kusha Thanks for addressing the comments. Please file issue/PR to keep track of pending items. I will merge it in. Thanks for the contribution! This will be very useful!

anakrish (approved):

PR #672, opened 2026-04-09
feat(compiler): FFI and C# bindings for RVM, incl. host-await built-ins

kusha:
## Summary

Exposes the registered host-await builtins, `Program` compilation, and RVM runtime accessors through the FFI layer and C# bindings. This is the companion to #667 (registered host-await builtins in the compiler/VM), making the feature usable from C# consumers.

> Rebased onto the updated #667. The multi-identifier `SetHostAwaitResponses` API (below) supersedes the earlier single-identifier form and addresses the review note about `regorus_rvm_set_host_await_responses` clearing other identifiers' queues.

## Motivation

The previous PR added registered host-await builtins to the Rust compiler and VM. However, the FFI boundary and C# bindings only exposed the raw `__builtin_host_await` path. This PR bridges the gap so C# consumers can:

1. **Register host-await builtins at compile time** - pass builtin names to `Program.CompileFromModules`, which emits `HostAwait` instructions directly.
2. **Pre-load responses for run-to-completion mode** - call `Rvm.SetHostAwaitResponses` to queue responses (for any number of identifiers) before execution.
3. **Inspect suspension state** - call `Rvm.GetHostAwaitIdentifier()` and `Rvm.GetHostAwaitArgument()` to determine which builtin suspended and with what argument.

### Rust VM (`src/rvm/vm/machine.rs`)

- `get_host_await_argument()` and `get_host_await_identifier()` - `const fn` accessors that return the argument/identifier when the VM is in a HostAwait-suspended state.

### FFI (`bindings/ffi/src/rvm.rs`)

- `RegorusHostAwaitBuiltin` - `#[repr(C)]` struct carrying a builtin name. The argument count is fixed to 1 by the compiler, so it is not part of the FFI surface.
- `regorus_program_compile_from_modules_with_host_await` - compile from modules with registered host-await builtins; the non-host-await path is treated as the empty-builtins case.
- `regorus_rvm_set_host_await_responses` - pre-load responses for run-to-completion mode. Accepts an array of `RegorusHostAwaitResponseSet` (one per identifier), so multiple identifiers are configured in a single call.
- `regorus_rvm_get_host_await_argument` / `regorus_rvm_get_host_await_identifier` - JSON accessors for suspension state.

### C# Bindings (`bindings/csharp/Regorus/`)

- `HostAwaitBuiltin` - readonly struct wrapping a builtin name (single-argument constructor; the arg count is fixed to 1 at the compiler level).
- `ExecutionMode` - enum (`RunToCompletion`, `Suspendable`).
- `Program.CompileFromModules` - overload accepting `IReadOnlyList<HostAwaitBuiltin>`.
- `Rvm.SetHostAwaitResponses(IReadOnlyDictionary<string, IReadOnlyList<string>>)` - queue JSON responses for any number of identifiers in one call. Replaces the entire response configuration.
- `Rvm.GetHostAwaitArgument()` / `GetHostAwaitIdentifier()` - read suspension state.
- `ModuleMarshalling` - `PinnedUtf8Strings`, `PinnedHostAwaitBuiltins`, and `PinnedHostAwaitResponseSets` helpers for safe FFI marshalling.

### Documentation

- **`bindings/csharp/README.md`** - suspendable and run-to-completion examples with registered builtins.
- **`bindings/csharp/API.md`** - `Program`, `Rvm`, `HostAwaitBuiltin`, `ExecutionMode` API reference.
- **`docs/rvm/vm-runtime.md`** - the new VM accessors.

### Tests (`bindings/csharp/Regorus.Tests/RvmProgramTests.cs`)

- `RegisteredHostAwait_Suspendable_SuspendAndResume` - registers `get_account`, suspends, verifies identifier and argument, resumes with a response.
- `RegisteredHostAwait_RunToCompletion_WithPreloadedResponses` - registers `translate`, pre-loads a response, verifies end-to-end execution.
- `RegisteredHostAwait_RunToCompletion_MultipleIdentifiersInSingleCall` - preloads responses for two identifiers in one `SetHostAwaitResponses` call.
- `RegisteredHostAwait_RunToCompletion_FailsWhenResponseQueueExhausted` - missing response surfaces as an error.
- `RegisteredHostAwait_GetAccessorsReturnNullWhenVmIsNotSuspended` - accessors return null outside a suspension.
- `RegisteredHostAwait_CompileRejectsEmptyOrWhitespaceName` / `CompileRejectsDuplicateRegistration` / `CompileRejectsReservedName` - registration validation errors surface through the bindings.

## Notes

- **`SetHostAwaitResponses` replaces the full response set**: it configures responses for all supplied identifiers in a single call and replaces any previously configured responses. This mirrors the underlying `RegoVM::set_host_await_responses` replace-all semantics (pre-existing upstream behavior).

- **`CompileFromEngine` does not yet support host-await builtins**: Only `CompileFromModules` accepts `HostAwaitBuiltin[]`. This is a scoping choice to keep the PR smaller - there is no technical constraint preventing it. The engine-based path can be extended in a follow-up if needed.

### DIFF
--- a/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
+++ b/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
@@ -116,4 +116,78 @@ allow if {
         var resumed = vm.Resume("{\"tier\":\"gold\"}");
         Assert.AreEqual("true", resumed, "expected allow=true after resume");
     }
+
+    private const string GetAccountPolicy = """
+package demo
+import rego.v1
+
+default allow := false
+
+allow if {
+  account := get_account({"id": input.account_id})
+  account.status == "active"
+}
+""";
+
+    [TestMethod]
+    public void RegisteredHostAwait_Suspendable_SuspendAndResume()
+    {
+        var modules = new[] { new PolicyModule("account.rego", GetAccountPolicy) };
+        var entryPoints = new[] { "data.demo.allow" };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("get_account", 1) };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.Suspendable);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"account_id\": \"acct-42\"}");
+
+        // Execute — should suspend on get_account()
+        vm.Execute();
+
+        // Verify we're suspended due to HostAwait with identifier "get_account"
+        var identifier = vm.GetHostAwaitIdentifier();
+        Assert.AreEqual("\"get_account\"", identifier, "expected identifier to be get_account");
+
+        var argument = vm.GetHostAwaitArgument();
+        Assert.IsNotNull(argument, "expected non-null argument");
+        StringAssert.Contains(argument!, "acct-42", "expected account_id in argument");
+
+        // Resume with an account response
+        var result = vm.Resume("{\"status\": \"active\", \"name\": \"Alice\"}");
+        Assert.AreEqual("true", result, "expected allow=true after resume");
+    }
+
+    private const string TranslatePolicy = """
+package demo
+import rego.v1
+
+default greeting := "unknown"
+
+greeting := msg if {
+  msg := translate(input.lang)
+}
+""";
+
+    [TestMethod]
+    public void RegisteredHostAwait_RunToCompletion_WithPreloadedResponses()
+    {
+        var modules = new[] { new PolicyModule("translate.rego", TranslatePolicy) };
+        var entryPoints = new[] { "data.demo.greeting" };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate", 1) };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.RunToCompletion);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"lang\": \"es\"}");
+
+        // Pre-load a response for translate
+        vm.SetHostAwaitResponses("translate", new[] { "\"hola\"" });
+
+        // Execute — translate returns "hola"
+        var result = vm.Execute();
+        Assert.AreEqual("\"hola\"", result, "expected greeting=hola");
+    }
+
 }
--- a/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
+++ b/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
@@ -237,6 +237,116 @@ greeting := msg if {
             "expected compilation to reject reserved __builtin_host_await identifier");
     }
 
+    public static IEnumerable<object[]> EmbeddedNulIdentifierScenarios()
+    {
+        // Identifiers (registration name + response key) are raw C strings with
+        // no downstream parser to catch truncation, so an embedded NUL would
+        // silently mis-route. Both write-points reject it.
+        yield return new object[]
+        {
+            "builtin name",
+            (Action)(() => _ = new HostAwaitBuiltin("bad\0name")),
+        };
+        yield return new object[]
+        {
+            "response identifier",
+            (Action)(() =>
+            {
+                using var vm = new Rvm();
+                vm.SetHostAwaitResponses(new Dictionary<string, IReadOnlyList<string>>
+                {
+                    ["bad\0id"] = new[] { "\"v\"" },
+                });
+            }),
+        };
+    }
+
+    // A raw NUL in an identifier is rejected loudly at the public surface,
+    // before Rust's CStr can silently truncate it.
+    [DataTestMethod]
+    [DynamicData(nameof(EmbeddedNulIdentifierScenarios), DynamicDataSourceType.Method)]
+    public void RegisteredHostAwait_RejectsEmbeddedNulInIdentifier(string description, Action action)
+    {
+        Assert.ThrowsException<ArgumentException>(
+            action,
+            $"expected embedded NUL in {description} to be rejected before crossing the FFI boundary");
+    }
+
+    // A JSON-escaped null (\u0000) is six ASCII chars, not a raw NUL, so it
+    // round-trips faithfully. Response values are not NUL-validated (client
+    // responsibility, consistent with other JSON inputs).
+    [TestMethod]
+    public void RegisteredHostAwait_ResponseValueWithEscapedNul_RoundTrips()
+    {
+        var modules = new[] { new PolicyModule("translate.rego", TranslatePolicy) };
+        var entryPoints = new[] { "data.demo.greeting" };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate") };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.RunToCompletion);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"lang\": \"es\"}");
+
+        vm.SetHostAwaitResponses(new Dictionary<string, IReadOnlyList<string>>
+        {
+            ["translate"] = new[] { "\"a\\u0000b\"" },
+        });
+
+        var result = vm.Execute();
+
+        Assert.AreEqual("\"a\\u0000b\"", result);
+    }
+
+    // Same escaped-null round-trip on the resume path (customer-controlled JSON,
+    // not NUL-validated).
+    [TestMethod]
+    public void RegisteredHostAwait_ResumeValueWithEscapedNul_RoundTrips()
+    {
+        var modules = new[] { new PolicyModule("translate.rego", TranslatePolicy) };
+        var entryPoints = new[] { "data.demo.greeting" };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate") };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.Suspendable);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"lang\": \"es\"}");
+
+        vm.Execute();
+        Assert.AreEqual("translate", vm.GetHostAwaitIdentifier());
+
+        var result = vm.Resume("\"a\\u0000b\"");
+
+        Assert.AreEqual("\"a\\u0000b\"", result);
+    }
+
+    // Failing scenario: a raw NUL in the resume value is not validated, but
+    // Rust's CStr truncates at it and the truncated text ("a) is invalid JSON,
+    // so it surfaces as a loud parse error. (A raw NUL that truncates to *valid*
+    // JSON — e.g. "123\0456" -> 123 — is silently accepted, the accepted
+    // binding-wide value contract shared with AddDataJson/SetInputJson.)
+    [TestMethod]
+    public void RegisteredHostAwait_ResumeValueWithRawNul_ThrowsFromParse()
+    {
+        var modules = new[] { new PolicyModule("translate.rego", TranslatePolicy) };
+        var entryPoints = new[] { "data.demo.greeting" };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate") };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.Suspendable);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"lang\": \"es\"}");
+
+        vm.Execute();
+        Assert.AreEqual("translate", vm.GetHostAwaitIdentifier());
+
+        Assert.ThrowsException<InvalidOperationException>(
+            () => vm.Resume("\"a\0b\""),
+            "expected a raw NUL that truncates to invalid JSON to surface as a parse error");
+    }
+
     [TestMethod]
     public void RegisteredHostAwait_GetAccessorsReturnNullWhenVmIsNotSuspended()
     {
--- a/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
+++ b/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Regorus.Tests;
@@ -134,7 +135,7 @@ allow if {
     {
         var modules = new[] { new PolicyModule("account.rego", GetAccountPolicy) };
         var entryPoints = new[] { "data.demo.allow" };
-        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("get_account", 1) };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("get_account") };
 
         using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
         using var vm = new Rvm();
@@ -145,7 +146,9 @@ allow if {
         // Execute — should suspend on get_account()
         vm.Execute();
 
-        // Verify we're suspended due to HostAwait with identifier "get_account"
+        // Verify we're suspended due to HostAwait with identifier "get_account".
+        // GetHostAwaitIdentifier returns the JSON-encoded Value, so the identifier
+        // string itself includes the surrounding JSON quotes.
         var identifier = vm.GetHostAwaitIdentifier();
         Assert.AreEqual("\"get_account\"", identifier, "expected identifier to be get_account");
 
@@ -174,7 +177,7 @@ greeting := msg if {
     {
         var modules = new[] { new PolicyModule("translate.rego", TranslatePolicy) };
         var entryPoints = new[] { "data.demo.greeting" };
-        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate", 1) };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate") };
 
         using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
         using var vm = new Rvm();
@@ -183,11 +186,161 @@ greeting := msg if {
         vm.SetInputJson("{\"lang\": \"es\"}");
 
         // Pre-load a response for translate
-        vm.SetHostAwaitResponses("translate", new[] { "\"hola\"" });
+        vm.SetHostAwaitResponses(new Dictionary<string, IReadOnlyList<string>>
+        {
+            ["translate"] = new[] { "\"hola\"" },
+        });
 
         // Execute — translate returns "hola"
         var result = vm.Execute();
         Assert.AreEqual("\"hola\"", result, "expected greeting=hola");
+    }
+
+    [TestMethod]
+    public void RegisteredHostAwait_CompileRejectsEmptyOrWhitespaceName()
+    {
+        var modules = new[] { new PolicyModule("noop.rego", "package demo\nallow := true\n") };
+        var entryPoints = new[] { "data.demo.allow" };
+
+        foreach (var badName in new[] { "", "   ", "\t" })
+        {
+            var builtins = new[] { new HostAwaitBuiltin(badName) };
+            Assert.ThrowsException<InvalidOperationException>(
+                () => Program.CompileFromModules("{}", modules, entryPoints, builtins),
+                $"expected compilation to reject empty/whitespace name '{badName}'");
+        }
+    }
+
+    [TestMethod]
+    public void RegisteredHostAwait_CompileRejectsDuplicateRegistration()
+    {
+        var modules = new[] { new PolicyModule("noop.rego", "package demo\nallow := true\n") };
+        var entryPoints = new[] { "data.demo.allow" };
+        var builtins = new[]
+        {
+            new HostAwaitBuiltin("translate"),
+            new HostAwaitBuiltin("translate"),
+        };
+
+        Assert.ThrowsException<InvalidOperationException>(
+            () => Program.CompileFromModules("{}", modules, entryPoints, builtins),
+            "expected compilation to reject duplicate registration");
+    }
+
+    [TestMethod]
+    public void RegisteredHostAwait_CompileRejectsReservedName()
+    {
+        var modules = new[] { new PolicyModule("noop.rego", "package demo\nallow := true\n") };
+        var entryPoints = new[] { "data.demo.allow" };
+        var builtins = new[] { new HostAwaitBuiltin("__builtin_host_await") };
+
+        Assert.ThrowsException<InvalidOperationException>(
+            () => Program.CompileFromModules("{}", modules, entryPoints, builtins),
+            "expected compilation to reject reserved __builtin_host_await identifier");
+    }
+
+    [TestMethod]
+    public void RegisteredHostAwait_GetAccessorsReturnNullWhenVmIsNotSuspended()
+    {
+        var modules = new[] { new PolicyModule("translate.rego", TranslatePolicy) };
+        var entryPoints = new[] { "data.demo.greeting" };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate") };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.RunToCompletion);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"lang\": \"es\"}");
+        vm.SetHostAwaitResponses(new Dictionary<string, IReadOnlyList<string>>
+        {
+            ["translate"] = new[] { "\"hola\"" },
+        });
+        vm.Execute();
+
+        // After run-to-completion completes successfully, the VM is no longer suspended.
+        Assert.IsNull(vm.GetHostAwaitArgument(), "expected null argument when VM is not suspended");
+        Assert.IsNull(vm.GetHostAwaitIdentifier(), "expected null identifier when VM is not suspended");
+    }
+
+    private const string TranslateNoDefaultPolicy = """
+package demo
+import rego.v1
+
+# No default — if translate() can't produce a value, the entry point
+# evaluation propagates the error to the caller.
+result := translate(input.lang)
+""";
+
+    [TestMethod]
+    public void RegisteredHostAwait_RunToCompletion_FailsWhenResponseQueueExhausted()
+    {
+        var modules = new[] { new PolicyModule("translate.rego", TranslateNoDefaultPolicy) };
+        var entryPoints = new[] { "data.demo.result" };
+        var hostAwaitBuiltins = new[] { new HostAwaitBuiltin("translate") };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.RunToCompletion);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"lang\": \"es\"}");
+
+        // No responses pre-loaded — translate has nothing to return.
+        // Document the actual behavior: in run-to-completion mode the
+        // missing-response error fails the rule body silently rather than
+        // surfacing as an exception, so Execute() returns the literal
+        // string `"<undefined>"` for an entry point that produced no value.
+        // Asserting the exact return value locks this contract so any
+        // future change (e.g. propagating an exception) shows up as a
+        // test failure that has to be explicitly re-acknowledged.
+        var actual = vm.Execute();
+        Assert.AreEqual(
+            "\"<undefined>\"",
+            actual,
+            "expected `\"<undefined>\"` when the response queue is exhausted");
+    }
+
+    private const string MultiAwaitPolicy = """
+package demo
+import rego.v1
+
+default greeting := "unknown"
+
+greeting := combined if {
+  hello := translate(input.lang)
+  user := lookup_user({"id": input.user_id})
+  combined := sprintf("%s %s", [hello, user.name])
+}
+""";
+
+    [TestMethod]
+    public void RegisteredHostAwait_RunToCompletion_MultipleIdentifiersInSingleCall()
+    {
+        var modules = new[] { new PolicyModule("multi.rego", MultiAwaitPolicy) };
+        var entryPoints = new[] { "data.demo.greeting" };
+        var hostAwaitBuiltins = new[]
+        {
+            new HostAwaitBuiltin("translate"),
+            new HostAwaitBuiltin("lookup_user"),
+        };
+
+        using var program = Program.CompileFromModules("{}", modules, entryPoints, hostAwaitBuiltins);
+        using var vm = new Rvm();
+        vm.SetExecutionMode(ExecutionMode.RunToCompletion);
+        vm.LoadProgram(program);
+        vm.SetInputJson("{\"lang\": \"es\", \"user_id\": \"u1\"}");
+
+        // Pre-load responses for BOTH identifiers in a single call.
+        // The new IReadOnlyDictionary API atomically replaces ALL prior
+        // responses, so this single call must carry every identifier the
+        // policy may invoke during this run.
+        vm.SetHostAwaitResponses(new Dictionary<string, IReadOnlyList<string>>
+        {
+            ["translate"] = new[] { "\"hola\"" },
+            ["lookup_user"] = new[] { "{\"name\": \"Alice\"}" },
+        });
+
+        var result = vm.Execute();
+        Assert.AreEqual("\"hola Alice\"", result, "expected combined greeting from both responses");
     }
 
 }
--- a/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
+++ b/bindings/csharp/Regorus.Tests/RvmProgramTests.cs
@@ -147,10 +147,8 @@ allow if {
         vm.Execute();
 
         // Verify we're suspended due to HostAwait with identifier "get_account".
-        // GetHostAwaitIdentifier returns the JSON-encoded Value, so the identifier
-        // string itself includes the surrounding JSON quotes.
         var identifier = vm.GetHostAwaitIdentifier();
-        Assert.AreEqual("\"get_account\"", identifier, "expected identifier to be get_account");
+        Assert.AreEqual("get_account", identifier, "expected identifier to be get_account");
 
         var argument = vm.GetHostAwaitArgument();
         Assert.IsNotNull(argument, "expected non-null argument");
--- a/bindings/csharp/Regorus/ModuleMarshalling.cs
+++ b/bindings/csharp/Regorus/ModuleMarshalling.cs
@@ -45,12 +45,12 @@ namespace Regorus.Internal
             }
         }
 
-        internal sealed class PinnedEntryPoints : IDisposable
+        internal sealed class PinnedUtf8Strings : IDisposable
         {
             private readonly List<Utf8Marshaller.PinnedUtf8> _pins;
             private bool _disposed;
 
-            internal PinnedEntryPoints(IntPtr[] buffer, int length, List<Utf8Marshaller.PinnedUtf8> pins)
+            internal PinnedUtf8Strings(IntPtr[] buffer, int length, List<Utf8Marshaller.PinnedUtf8> pins)
             {
                 Buffer = buffer;
                 Length = length;
@@ -119,14 +119,14 @@ namespace Regorus.Internal
             }
         }
 
-        internal static PinnedEntryPoints PinEntryPoints(IReadOnlyList<string> entryPoints)
+        internal static PinnedUtf8Strings PinUtf8Strings(IReadOnlyList<string> values)
         {
-            if (entryPoints is null)
+            if (values is null)
             {
-                throw new ArgumentNullException(nameof(entryPoints));
+                throw new ArgumentNullException(nameof(values));
             }
 
-            var count = entryPoints.Count;
+            var count = values.Count;
             var buffer = ArrayPool<IntPtr>.Shared.Rent(count);
             var pins = new List<Utf8Marshaller.PinnedUtf8>(count);
 
@@ -134,12 +134,12 @@ namespace Regorus.Internal
             {
                 for (int i = 0; i < count; i++)
                 {
-                    var entryPinned = Utf8Marshaller.Pin(entryPoints[i]);
-                    pins.Add(entryPinned);
-                    buffer[i] = (IntPtr)entryPinned.Pointer;
+                    var pinned = Utf8Marshaller.Pin(values[i]);
+                    pins.Add(pinned);
+                    buffer[i] = (IntPtr)pinned.Pointer;
                 }
 
-                return new PinnedEntryPoints(buffer, count, pins);
+                return new PinnedUtf8Strings(buffer, count, pins);
             }
             catch
             {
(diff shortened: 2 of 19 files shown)


tests/rvm/rego/cases/registered_host_await.yaml
```diff
@@ -0,0 +1,240 @@
...
+      - name: kv_store
+        arg_count: 2
+    modules:
+      - |
+        package demo
+        import rego.v1
+
+        result := kv_store(input.key, input.value)
+    query: data.demo.result
+    # Registration panics because arg_count must be 1.
```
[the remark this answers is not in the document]
kusha: Fixed — the comment now says the registration fails with a compilation error rather than "panics".

bindings/csharp/Regorus/ModuleMarshalling.cs
```diff
@@ -152,5 +152,77 @@ internal static PinnedEntryPoints PinEntryPoints(IReadOnlyList<string> entryPoin
...
+            {
+                for (int i = 0; i < count; i++)
+                {
+                    var namePinned = Utf8Marshaller.Pin(builtins[i].Name);
+                    pins.Add(namePinned);
+
+                    buffer[i] = new RegorusHostAwaitBuiltin
+                    {
+                        name = namePinned.Pointer,
+                        arg_count = (UIntPtr)builtins[i].ArgCount,
```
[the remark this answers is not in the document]
kusha: Resolved by removing the field:  HostAwaitBuiltin  no longer carries an argument count (it's fixed to 1 by the compiler), so there's no  int → UIntPtr  cast anymore.

bindings/csharp/Regorus/Rvm.cs
```diff
@@ -195,6 +195,63 @@ public void SetExecutionMode(ExecutionMode mode)
...
+                return CheckAndDropResult(API.regorus_rvm_get_host_await_identifier((RegorusRvm*)vmPtr));
+            });
+        }
+
+        /// <summary>
+        /// Pre-load HostAwait responses for run-to-completion mode.
+        /// Clears any previously configured responses, then queues the
+        /// provided values for the given identifier.
+        /// </summary>
+        /// <param name="identifier">The builtin identifier.</param>
```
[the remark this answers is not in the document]
kusha: Addressed in the rebase: the single-identifier method is replaced by SetHostAwaitResponses(IReadOnlyDictionary<...>) , which configures all identifiers at once. The XML doc now states it replaces the entire response set.

bindings/ffi/src/rvm.rs
```diff
@@ -604,3 +627,157 @@ fn convert_c_modules_to_rust(
...
+                    }
+                    let ptr = *values_json.add(i);
+                    let json_str = from_c_str(ptr)?;
+                    let val = Value::from_json_str(&json_str)?;
+                    values.push_back(val);
+                }
+            }
+
+            guard.set_host_await_responses(core::iter::once((id_value, values)));
+            Ok(())
```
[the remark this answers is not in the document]
kusha: Addressed:  regorus_rvm_set_host_await_responses  now takes an array of  RegorusHostAwaitResponseSet  (and the C# side takes  IReadOnlyDictionary<string, IReadOnlyList<string>> ), so all identifiers are configured in a single call — no more dropping previously-set identifiers. The underlying RegoVM::set_host_await_responses  replace-all semantics are pre-existing behavior; changing those to per-identifier clearing would be a separate VM-API change, out of scope here.

anakrish:
@</Copilot> review this PR using all the review skills in the repo. Use a separate agent for each skill.

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +669,190 @@ fn convert_c_modules_to_rust(
...
+/// * `response_sets_len` - Number of entries in `response_sets`
+#[no_mangle]
+pub extern "C" fn regorus_rvm_set_host_await_responses(
+    vm: *mut RegorusRvm,
+    response_sets: *const RegorusHostAwaitResponseSet,
+    response_sets_len: usize,
+) -> RegorusResult {
+    with_unwind_guard(|| {
+        to_regorus_result(|| -> Result<()> {
+            let vm = to_ref(vm)?;
```
anakrish: These new host-await entry points reach for the VM handle with `to_ref(vm)?`, but every other RVM method in this file (e.g. the run/step/result functions above) goes through `to_shared_ref(vm as *const RegorusRvm)?` and *then* takes the lock via `try_read`/`try_write`. That ordering isn't accidental: `to_shared_ref` is what lets the lock arbitrate concurrent access, whereas `to_ref` manufactures an exclusive `&mut RegorusRvm` straight from a caller-owned handle. If two foreign threads call into the same `vm` here, both can materialize that `&mut` before either lock is acquired, which is exactly the aliasing hazard the rest of the file is structured to avoid.

Could we switch these three (`set_host_await_responses`, `get_host_await_argument`, `get_host_await_identifier`) to `to_shared_ref(...)` followed by `try_write`/`try_read`, to match the established pattern?

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +669,190 @@ fn convert_c_modules_to_rust(
...
+            let mut guard = vm.try_write()?;
+
+            if response_sets.is_null() && response_sets_len > 0 {
+                return Err(anyhow!("null response_sets pointer"));
+            }
+
+            let mut all = Vec::with_capacity(response_sets_len);
```
anakrish: These `Vec::with_capacity(response_sets_len)` / `VecDeque::with_capacity(set.values_len)` (and the `Vec::with_capacity(len)` further down) reserve straight from caller-controlled `usize` values. A hostile or buggy caller passing something near `usize::MAX` will trip a capacity-overflow panic *inside* the unwind guard, which for this engine means process-wide poisoning rather than a clean error return.

I'll flag that this isn't new to your PR — the existing `compile_from_modules` paths in this file already use `with_capacity(modules_len)` the same way — so I don't think it should block the PR on its own. But since we're adding several more instances of the pattern, it'd be a good moment to switch to fallible reservation (`try_reserve`) or validate against a sane maximum before allocating, and return an FFI error instead of risking a panic.

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +669,190 @@ fn convert_c_modules_to_rust(
...
+            for i in 0..response_sets_len {
+                // SAFETY: caller guarantees `response_sets` points to a
+                // contiguous array of `response_sets_len` `RegorusHostAwaitResponseSet`
+                // values, and the inner pointers reference valid C strings.
+                let set = unsafe { &*response_sets.add(i) };
+
+                let id_str = from_c_str(set.identifier)
+                    .map_err(|e| anyhow!("invalid identifier in response set at index {i}: {e}"))?;
+                let id_value = Value::String(id_str.into());
```
anakrish, line 1201: Worth tracing the identifier round-trip here. On the way in we wrap the C string as `Value::String(id_str.into())`, but the getter below (`get_host_await_identifier`) hands the identifier back as `to_json_str()`, i.e. a JSON-encoded value. So a host that reads the current identifier and feeds it straight back gets `Value::String("\"get_account\"")` — the quotes become part of the string and it no longer matches.

It also means a HostAwait identifier that isn't a string (say an integer `42` baked into the policy) can never be matched from the FFI side at all, since everything coming through here is coerced to `Value::String`. If identifiers are intended to be string-only, it'd be clearer to document and enforce that and have the getter return the raw string; otherwise we probably want an `identifier_json` in and a parse back to `Value`.

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +669,190 @@ fn convert_c_modules_to_rust(
...
+/// Returns the argument value if the VM is suspended due to a HostAwait instruction,
+/// or None if the VM is not in a HostAwait-suspended state.
+#[no_mangle]
+pub extern "C" fn regorus_rvm_get_host_await_argument(vm: *mut RegorusRvm) -> RegorusResult {
+    with_unwind_guard(|| {
+        let output = || -> Result<Option<String>> {
+            let vm = to_ref(vm)?;
+            let guard = vm.try_read()?;
+            match guard.get_host_await_argument() {
+                Some(arg) => Ok(Some(arg.to_json_str()?)),
```
anakrish, line 1251: `to_json_str()` is lossy for the Rego-only `Value` variants. If a policy hands a `Set` to HostAwait, it serializes here indistinguishably from an `Array`, and when the host resumes with that JSON it comes back as an `Array` — so a downstream `is_set(resp)` flips to false. `Undefined` has the same problem. For most payloads this is fine, but it's a silent shape change at the boundary. Either we should reject the non-JSON-representable variants explicitly at this boundary (so the failure is loud), or expose a typed marshalling path that preserves them.

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +669,190 @@ fn convert_c_modules_to_rust(
...
 }
+
+/// A registered host-awaitable builtin passed via FFI.
+///
+/// The argument count is currently fixed to 1 by the compiler (see
+/// `Compiler::register_host_await_builtin`), so it is not exposed at the
+/// FFI boundary. The struct exists as a stable layout to allow future
+/// expansion (e.g. an explicit `arg_count` field) without breaking ABI
+/// when callers pin a fixed-size array of these.
+#[repr(C)]
```
anakrish, line 1026: Since these `#[repr(C)]` structs get passed as C arrays, their size is baked into the caller's stride arithmetic. The moment we want to add a field later (an arity, a flags word, etc.) `sizeof(RegorusHostAwaitBuiltin)` changes and any client compiled against the old layout walks the array incorrectly — a silent ABI break rather than a link error. If we expect these to grow, it's cheap to add a `struct_size`/`version` (or a couple of reserved fields) now so future additions stay backward-compatible.

src/languages/rego/compiler/function_calls.rs
```diff
@@ -187,8 +239,26 @@ impl<'a> Compiler<'a> {
...
+        // We deliberately match against `original_fcn_path` only, not
+        // `full_fcn_path`. Registration intercepts the *unqualified* call
+        // form (e.g. `lookup(x)` inside the policy's own package). A
+        // package-qualified call like `data.other.lookup(x)` is left to
+        // resolve through the normal user-defined / builtin path, so a
+        // registered name does not leak into unrelated packages that
+        // happen to expose a rule with the same identifier. This is
+        // documented on `register_host_await_builtin`; the
+        // `registered_host_await.yaml` suite pins the behavior.
+        if self.host_await_builtins.contains_key(original_fcn_path) {
```
anakrish, line 293: This lookup keys off `original_fcn_path`, so a registration like `register_host_await_builtin("data.other.lookup", 1)` will happily intercept a real package-qualified call `data.other.lookup(x)` — even when that name resolves to a user-defined rule/function. That contradicts the documented contract that package-qualified calls aren't intercepted, and it's a quiet way to shadow a policy's own function with a host-await.

The registration validation already rejects reserved names, whitespace and bad arity, so the natural place to close this is there: reject names containing `.` (host-await builtins are unqualified), or, if qualified interception is actually intended, drop the "not intercepted" wording from the docs and add a test that pins the behavior.

src/languages/rego/compiler/rules.rs
```diff
@@ -234,8 +234,20 @@ impl<'a> Compiler<'a> {
...
         entry_points: &[&str],
+    ) -> Result<Arc<Program>> {
+        Self::compile_from_policy_with_host_await(policy, entry_points, &[])
+    }
+
+    /// Compile from a CompiledPolicy to RVM Program with registered host-awaitable builtins.
+    pub fn compile_from_policy_with_host_await(
+        policy: &CompiledPolicy,
+        entry_points: &[&str],
+        host_await_builtins: &[(&str, usize)],
```
anakrish, line 245: Minor design heads-up on the public signature: `&[(&str, usize)]` freezes the registration metadata to exactly "name + arity". The first time we want a third per-builtin attribute we're forced into either a breaking signature change or a parallel `_v2` entry point. A small `HostAwaitBuiltinSpec` (with a constructor/builder) would let this grow additively. Not blocking, just easier to live with long-term while the API is still young.

bindings/csharp/Regorus/Program.cs
```diff
@@ -123,16 +135,28 @@ public static Program CompileFromEngine(Engine engine, IReadOnlyList<string> ent
...
-            return engine.UseHandleForInterop(enginePtr =>
+            return Utf8Marshaller.WithUtf8(dataJson, dataPtr =>
             {
+                fixed (RegorusPolicyModule* modulesPtr = pinnedModules.Buffer)
                 fixed (IntPtr* entryPtr = pinnedEntryPoints.Buffer)
+                fixed (RegorusHostAwaitBuiltin* builtinsPtr = pinnedBuiltins.Buffer)
                 {
-                    var result = API.regorus_engine_compile_program_with_entrypoints(
-                        (RegorusEngine*)enginePtr,
+                    var result = API.regorus_program_compile_from_modules_with_host_await(
```
anakrish: Heads-up on a compatibility regression: the previous overloads used to call `regorus_program_compile_from_modules`, and now every path routes through `regorus_program_compile_from_modules_with_host_await`, including the no-builtins case. That means an existing app that updates only the managed package against an older native library will hit an `EntryPointNotFoundException` at runtime, even though it never asked for host-await.

Could we keep calling the original `regorus_program_compile_from_modules` when `hostAwaitBuiltins` is null/empty, and only reach for the new symbol when there's actually something to register? That keeps the old code path binary-compatible.

bindings/csharp/Regorus/Program.cs
```diff
@@ -51,9 +51,32 @@ public static Program CompileFromModules(string dataJson, IEnumerable<PolicyModu
         /// </summary>
         public static Program CompileFromModules(string dataJson, IReadOnlyList<PolicyModule> modules, IReadOnlyList<string> entryPoints)
         {
-            if (modules is null)
+            return CompileFromModulesInner(dataJson, modules, entryPoints, hostAwaitBuiltins: null);
+        }        /// <summary>
```
anakrish: Tiny formatting bug: the `/// <summary>` starts on the same line as the closing `}`, so the compiler doesn't treat it as a doc comment for the following method — the XML docs for that member silently disappear. A newline before `/// <summary>` fixes it.

bindings/csharp/Regorus/ModuleMarshalling.cs
```diff
@@ -152,5 +153,220 @@ internal static PinnedEntryPoints PinEntryPoints(IReadOnlyList<string> entryPoin
...
+
+            var count = builtins.Count;
+            var buffer = ArrayPool<RegorusHostAwaitBuiltin>.Shared.Rent(count);
+            var pins = new List<Utf8Marshaller.PinnedUtf8>(count);
+
+            try
+            {
+                for (int i = 0; i < count; i++)
+                {
+                    var namePinned = Utf8Marshaller.Pin(builtins[i].Name);
```
anakrish, line 205: These `Utf8Marshaller.Pin(...)` calls (the builtin name here, and the response key/value pins further down) hand the string to the native side as a C string, but nothing rejects an embedded `\0`. Rust's `CStr` stops at the first NUL, so `new HostAwaitBuiltin("foo\0bar")` silently registers `"foo"`. It's an edge case, but it's the kind of silent truncation that's painful to debug, so it'd be worth rejecting embedded NULs in the public C# surface (or moving these to pointer+length strings).

anakrish (commented):
Automated multi-perspective review of the host-await FFI / C# binding work. Inline comments below, roughly highest-impact first.

Two items I couldn't attach inline because they fall outside this PR's diff, but that are worth a look while you're in here:
- **`RegoVM::set_host_await_responses` (machine.rs) accumulates response queues with no `enforce_limit()`/fallible reservation.** This is pre-existing code, not something you changed, but the new FFI surface makes it reachable with caller-supplied volume, so it may be worth hardening as a follow-up.
- The **`Vec::with_capacity(caller_len)` panic-on-huge-length** pattern I flagged in `rvm.rs` already exists in the older `compile_from_modules` paths too — a fallible-reservation pass across all of them would be a nice cleanup.

bindings/csharp/Regorus.Tests/RvmProgramTests.cs
```diff
@@ -116,4 +117,230 @@ public void Program_host_await_suspend_and_resume_succeeds()
...
+        // Document the actual behavior: in run-to-completion mode the
+        // missing-response error fails the rule body silently rather than
+        // surfacing as an exception, so Execute() returns the literal
+        // string `"<undefined>"` for an entry point that produced no value.
+        // Asserting the exact return value locks this contract so any
+        // future change (e.g. propagating an exception) shows up as a
+        // test failure that has to be explicitly re-acknowledged.
+        var actual = vm.Execute();
+        Assert.AreEqual(
+            "\"<undefined>\"",
```
anakrish, line 405: This is the one I'd most want a second opinion on. The test codifies that a run-to-completion HostAwait with no available response yields the string `"<undefined>"` rather than surfacing an error. Functionally that means a missing response is swallowed as a rule-body failure and propagates as plain undefined.

In a policy engine that's a fail-open shape: if the host *meant* to supply a response and didn't (misconfig, dropped lookup, etc.), the rule quietly evaluates to undefined instead of telling anyone something went wrong. I'd argue `HostAwaitResponseMissing` should propagate as an execution error through RVM → FFI → C#, and this assertion should expect an exception. If undefined-on-missing is genuinely the desired contract, it'd be good to spell out why in a comment, because the default reading is surprising.

bindings/ffi/src/rvm.rs
kusha: Switched all three (`set_host_await_responses`, `get_host_await_argument`, `get_host_await_identifier`) to `to_shared_ref(vm as *const RegorusRvm)?` followed by `try_write`/`try_read`, so the lock arbitrates concurrent access like the rest of the file. Left `regorus_rvm_drop` on `to_ref` - it needs exclusive ownership for `Box::from_raw` and takes no lock.

bindings/ffi/src/rvm.rs
kusha: Converted the three new host-await sites to `try_reserve`, so an absurd caller length returns a clean FFI error instead of a capacity-overflow panic under the unwind guard. As you noted, the pre-existing `compile_from_modules` sites share the pattern - I've left those out to keep this PR scoped and tracked them as a follow-up.

bindings/ffi/src/rvm.rs
kusha: Fixed - the getter now returns the raw identifier string instead of `to_json_str()`, so it feeds straight back into `set_host_await_responses`. Identifiers are string-only per the compiler, so I made the getter enforce that explicitly (rejecting a non-string identifier loudly) rather than the `identifier_json`-in route. Updated the C# docs and the round-trip test to expect the unquoted form.

bindings/ffi/src/rvm.rs
kusha: You're right it's lossy. We chose to make the convention explicit rather than reject: the argument uses the same `Value`→JSON encoding as evaluation results (which already emit `Set` as an array), so documenting it keeps one consistent mental model — and since policies are third-party-authored, rejecting a `Set` argument would turn a policy-authoring choice into a runtime failure.

The sharper `Undefined` half is a problem, let's proceed in the thread on `RvmProgramTests.cs`

bindings/ffi/src/rvm.rs
kusha: Added an explicit element-size parameter to both array-taking entry points; the native side validates it covers the current layout, then walks the array by the caller-supplied stride - so a caller built against a different struct layout indexes correctly or fails loud rather than corrupting memory. Applied to both new structs, with unit tests for the native-size, undersized-reject, and oversized (newer-caller) cases.

src/languages/rego/compiler/function_calls.rs
kusha: Closed it at registration: `register_host_await_builtin` now rejects names starting with `data.`. That's the namespace where user rules/functions live, so it removes the shadowing loophole while keeping the two intended uses - bare names register host functions, and builtin-namespace names like `time.parse_duration_ns` (not under `data.`) still override standard builtins. Added a negative test; the existing builtin-override case still passes.

src/languages/rego/compiler/rules.rs
kusha: Agreed it'll want a struct once there's a third attribute. Since this is an internal Rust signature (not the C ABI) with only a few in-tree callers, changing it later is a loud compile error rather than a silent break — so I'd rather introduce `HostAwaitBuiltinSpec` together with the arity work, when the real fields are known, instead of landing a 2-field wrapper now. Ok?

bindings/csharp/Regorus/Program.cs
kusha: Restored the dual-path routing: `CompileFromModulesInner` calls the original `regorus_program_compile_from_modules` when there are no builtins, and only reaches for `_with_host_await` when there's something to register. Keeps the no-host-await path binary-compatible with older native libraries.

bindings/csharp/Regorus/Program.cs
kusha: Fixed - added the newline so `CompileFromEngine`'s XML docs are recognized.

bindings/csharp/Regorus/ModuleMarshalling.cs
kusha: Fixed for the identifiers, which is where truncation actually mis-routes: an embedded NUL in the registration name or a response key would silently truncate with no downstream check, so both now reject it (shared `Utf8Marshaller.ThrowIfContainsNul`). I left response *values* unvalidated on purpose - they're payload backstopped by `from_json_str`, and a raw NUL in a JSON value is the same client-responsibility contract that already applies to `AddDataJson`/`SetInputJson` and module content, so I didn't want to single out host-await. Tests cover identifier rejection plus escaped `\u0000` round-trips on the preset and resume paths.

bindings/csharp/Regorus.Tests/RvmProgramTests.cs
kusha: I agree. To be clear on origin, the behavior is pre-existing from #667 — `HostAwaitResponseMissing` already isn't in `is_fatal_vm_error` (`src/rvm/vm/rules.rs`), so it's swallowed to undefined; this PR only added the test that pins it. It should propagate: a missing preloaded response is never a valid outcome, and the resource-limit errors are fatal precisely so `not builtin(...)` can't silently flip. Since fixing it is a core-RVM change across both execution paths and every binding (and it flips this test's contract), I'd prefer a separate focused PR. Would it be ok to solve in a follow-up?

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +676,347 @@ fn convert_c_modules_to_rust(
...
+            let guard = vm.try_read()?;
+            match guard.get_host_await_identifier() {
+                Some(Value::String(s)) => Ok(Some(s.as_ref().to_string())),
+                Some(_) => Err(anyhow!("host-await identifier must be a string")),
+                None => Ok(None),
+            }
+        }();
+
+        match output {
+            Ok(Some(identifier)) => RegorusResult::ok_string(identifier),
```
anakrish, line 1307: 🤖 **Automated review — actionable (correctness)**

`regorus_rvm_get_host_await_identifier` returns the identifier as a **raw** C string via `ok_string`. If a policy suspends with an identifier containing an interior NUL — e.g. `__builtin_host_await(x, "a\u0000b")` — `RegorusResult::ok_string` → `to_c_str` hits the `CString::new` failure branch in `common.rs` and **silently substitutes the literal `"binding error: failed to create c-style string"` while keeping `Ok`/`RegorusStatus::Ok`**. The caller receives a wrong identifier with no error signal and cannot route/resume the host-await.

Note the *argument* accessor is immune because it JSON-encodes the value; only this raw-string identifier path is affected (reachable via the raw `__builtin_host_await` string-literal form, not registered builtins).

Suggestion: reject interior-NUL identifiers explicitly, e.g.
```rust
Some(Value::String(s)) => {
    if s.contains('\0') {
        return Err(anyhow!("host-await identifier contains an embedded NUL"));
    }
    Ok(Some(s.as_ref().to_string()))
}
```
or JSON-encode the identifier so every Rego string is representable across the ABI.

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +676,347 @@ fn convert_c_modules_to_rust(
...
+            if response_sets.is_null() && response_sets_len > 0 {
+                return Err(anyhow!("null response_sets pointer"));
+            }
+
+            // `response_set_size` is `sizeof(RegorusHostAwaitResponseSet)` as the
+            // caller compiled it, which is also the array stride. Validate and use
+            // it so a caller built against a different struct layout still walks
+            // the array correctly.
+            let min_size = core::mem::size_of::<RegorusHostAwaitResponseSet>();
+            if response_sets_len > 0 && response_set_size < min_size {
```
anakrish: 🤖 **Automated review — ABI design (advisory)**

The stride/size design is documented as a "forward-compatible ABI," but the minimum-size gate uses `size_of::<CurrentStruct>()`. That only guarantees the *new-caller + old-native* direction (covered by `convert_builtins_oversized_stride_uses_caller_stride`). In the reverse direction — a **future native build that appends a field** — `min_size` grows, so an **older caller passing the old (smaller) size is rejected as an ABI mismatch**, contradicting the forward-compat claim.

Consider either (a) softening the doc comments to state the compatibility is one-directional, or (b) freezing a v1 prefix-size constant and defaulting missing trailing fields, or (c) adding a reserved/version field now. Same applies to the builtin-size check at ~L898.

bindings/ffi/src/rvm.rs
```diff
@@ -640,3 +676,347 @@ fn convert_c_modules_to_rust(
...
+            })?;
+            let base = response_sets as *const u8;
+            for i in 0..response_sets_len {
+                let offset = i.checked_mul(response_set_size).ok_or_else(|| {
+                    anyhow!("host-await response set array offset overflow at index {i}")
+                })?;
+                // SAFETY: caller guarantees `response_sets` points to a contiguous
+                // array of `response_sets_len` elements each `response_set_size`
+                // bytes wide, and the inner pointers reference valid C strings.
+                let set = unsafe { &*(base.add(offset) as *const RegorusHostAwaitResponseSet) };
```
anakrish, line 1197: 🤖 **Automated review — soundness hardening (advisory)**

`&*(base.add(offset) as *const RegorusHostAwaitResponseSet)` (and the sibling at ~L915 for `RegorusHostAwaitBuiltin`) forms a Rust reference from a caller-controlled offset. The size check validates stride but not **alignment** — a packed/misaligned C buffer could satisfy it while making the `&*` reference formation UB before any validation runs. In practice the C# caller passes an aligned pinned array, so this is within the documented `SAFETY` caller contract, but it's a real soundness edge for arbitrary C callers.

Hardening options: validate `base as usize % align_of::<T>() == 0` before deref, or read fields via `ptr::read_unaligned` instead of forming `&T`.

anakrish (commented):
## 🤖 Automated review summary

_This review was produced by automated code-review tooling (a fast multi-perspective pass and a multi-agent deep-review pass) and then human-calibrated. Treat findings as advisory — please verify before acting._

Overall the change is well-engineered: careful FFI pointer/stride validation, `try_reserve` everywhere, disciplined GCHandle/ArrayPool marshalling on the C# side, and a backward-compatible no-builtins path. Local validation passed: FFI unit tests (30), RVM rego YAML incl. `registered_host_await` (29), and the C# suite (72).

### Actionable
1. **Interior-NUL host-await identifier returns bogus success** (inline on `bindings/ffi/src/rvm.rs`). The one concrete correctness bug; both automated passes flagged it.

### Design / ABI considerations (advisory)
2. **Forward-compat ABI claim is one-directional** — the stride design only supports *new caller + old native*; an appended field would reject old callers (inline).
3. **Unaligned deref soundness** at the FFI stride walkers (inline).
4. **Explicit non-string identifiers** can't be preloaded/returned (string-only APIs). Errors cleanly; documented scoping limit.

### Intentional & already documented (noted, not blocking)
- Run-to-completion **missing-response → `"<undefined>"`** is pre-existing RVM semantics, and is locked in by `RvmProgramTests.cs`. A deep-review pass rated this "Critical"; we downgrade it since it is not introduced here — but flagging in case silent host-outage-as-deny is undesirable at the product level.
- **Response-value embedded-NUL** is deliberately not validated (keys are), consistent with the `AddDataJson`/`SetInputJson` binding-wide contract, and is covered by dedicated round-trip/parse-error tests.

### Unrelated to the feature
- `.github/workflows/*` Actions SHA bumps look like rebase artifacts; in `rust-clippy.yml` the SHA changed but the `# v3.29.11` comment did not — worth a consistency check.

**Recommendation:** request the interior-NUL identifier fix; the rest are advisory. Not an approval/blocker vote — posted as automated comments.

bindings/ffi/src/rvm.rs
kusha: Fixed in 6d46602 — rejects the interior NUL in a new
`host_await_identifier_string` helper rather than JSON-encoding, which would
break the raw-string round-trip into `set_host_await_responses`

bindings/ffi/src/rvm.rs
kusha: Let's do it as a follow up in next PR

bindings/ffi/src/rvm.rs
kusha: Let's do it as a follow up in next PR

anakrish:
### Blocking: validate caller-controlled FFI strides before pointer arithmetic

`bindings/ffi/src/rvm.rs:789-795` and `:933-938` use caller-supplied `response_set_size` / `host_await_builtin_size` as array strides:

```rust
let offset = i.checked_mul(response_set_size)?;
let set = unsafe { &*(base.add(offset) as *const RegorusHostAwaitResponseSet) };
```

The current check only verifies `stride >= size_of::<T>()`; it does not validate alignment or pointer-offset bounds. For example:

- `response_sets_len = 2`, `response_set_size = isize::MAX + 1` passes the minimum-size check, then reaches invalid pointer arithmetic for element 2.
- `response_set_size = size_of::<RegorusHostAwaitResponseSet>() + 1` makes element 2 misaligned before dereference.
- The same applies to `host_await_builtin_size`.

Please centralize checked-stride validation for both paths and reject, for nonzero lengths:

```rust
stride < size_of::<T>()
stride % align_of::<T>() != 0
stride > isize::MAX
(len - 1).checked_mul(stride) overflow
last_offset > isize::MAX - size_of::<T>()
```

Only call `base.add(offset)` after these checks. This preserves forward-compatible trailing fields while preventing malformed strides from reaching unsafe pointer operations.

Please add tests for undersized, misaligned, oversized, and multiplication-boundary strides for both structs.

anakrish:
### Blocking: validate caller-controlled FFI strides before pointer arithmetic

`bindings/ffi/src/rvm.rs:789-795` and `:933-938` use caller-supplied `response_set_size` / `host_await_builtin_size` as array strides:

```rust
let offset = i.checked_mul(response_set_size)?;
let set = unsafe { &*(base.add(offset) as *const RegorusHostAwaitResponseSet) };
```

The current check only verifies `stride >= size_of::<T>()`; it does not validate alignment or pointer-offset bounds. For example:

- `response_sets_len = 2`, `response_set_size = isize::MAX + 1` passes the minimum-size check, then reaches invalid pointer arithmetic for element 2.
- `response_set_size = size_of::<RegorusHostAwaitResponseSet>() + 1` makes element 2 misaligned before dereference.
- The same applies to `host_await_builtin_size`.

Please centralize checked-stride validation for both paths and reject, for nonzero lengths:

```rust
stride < size_of::<T>()
stride % align_of::<T>() != 0
stride > isize::MAX
(len - 1).checked_mul(stride) overflow
last_offset > isize::MAX - size_of::<T>()
```

Only call `base.add(offset)` after these checks. This preserves forward-compatible trailing fields while preventing malformed strides from reaching unsafe pointer operations. Please add tests for undersized, misaligned, oversized, and multiplication-boundary strides for both structs.

anakrish:
### Medium: Embedded NULs silently truncate response JSON and other C-string inputs

**Location:** `bindings/csharp/Regorus/ModuleMarshalling.cs:323-327`; `bindings/ffi/src/rvm.rs:814-821`; `bindings/csharp/Regorus/Utf8Marshaller.cs:105-145`

Response values and other C# C-string inputs (module IDs/content, data, input, entry-point names) are pinned as NUL-terminated strings without checking embedded NUL. `from_c_str` stops at the first NUL, so:

- Response `"123\0junk"` becomes JSON `"123"`; policy receives wrong value.
- Module content `"package p\nallow := true\0\nallow := false"` is evaluated only through the first prefix.

**Suggestion:** Reject embedded NUL in every response value via `Utf8Marshaller.ThrowIfContainsNul()`, or change the ABI to length-delimited UTF-8 buffers instead of C strings.

**Tests:** Existing tests cover escaped NUL and invalid JSON prefixes, but not raw-NUL embedded in valid JSON prefixes. Add table-driven coverage.

anakrish:
### Medium: C# context API unavailable when native lib built without azure_policy feature

**Location:** `bindings/ffi/src/rvm.rs:416-428`; `bindings/csharp/Regorus/NativeMethods.cs:186-187`; `bindings/csharp/Regorus/Rvm.cs:116-123`

The Rust export is gated behind `#[cfg(feature = "azure_policy")]`:

```rust
#[cfg(feature = "azure_policy")]
pub extern "C" fn regorus_rvm_set_context(...) { ... }
```

The C# wrapper imports and exposes `SetContextJson()` unconditionally:

```csharp
[DllImport(...)]
internal static extern RegorusResult regorus_rvm_set_context(...);

public void SetContextJson(string contextJson) { ... }
```

A native library built without `azure_policy` causes `EntryPointNotFoundException` when managed code calls `SetContextJson`.

**Suggestion:** Gate the C# wrapper consistently with the native export, or always export a stub returning an unsupported-feature error.

**Tests:** Add a no-`azure_policy` binding configuration check.

anakrish:
### Medium: Non-string host-await identifiers cannot be retrieved/preloaded through FFI

**Location:** `src/languages/rego/compiler/function_calls.rs:153-169`; `bindings/ffi/src/rvm.rs:866-878`

`__builtin_host_await(arg, id)` emits the second argument as an arbitrary `Value`, but the new FFI getter and response-preload APIs only accept string IDs:

```rust
id: arg_regs[1],  // Can be Number, Object, etc.
```

```rust
Some(_) => Err(anyhow!("host-await identifier must be a string")),
```

**Trace:** `__builtin_host_await("ping", 7)` stores `Value::Number(7)`; `GetHostAwaitIdentifier` returns an error and `SetHostAwaitResponses` cannot key the response by `7`.

**Suggestion:** Enforce string IDs at compile time via `type_check.rs`, or expose JSON/value-based identifier APIs and response keys to the FFI.

**Tests:** Add explicit non-string identifier compile/runtime coverage.

anakrish:
### Medium: Host-await response queues unbounded by resource limits

**Location:** `bindings/ffi/src/rvm.rs:781-823`

Caller-controlled `response_sets_len`, each `values_len`, and arbitrary JSON strings drive reservations and storage without `enforce_limit()` or any configured cap:

```rust
all.try_reserve(response_sets_len)?;
values.try_reserve(set.values_len)?;
for j in 0..set.values_len { Value::from_json_str(&json_str)?; }
```

This permits memory and CPU exhaustion in a process hosting the engine.

**Suggestion:** Enforce configured count/byte/value limits and charge parsing/storage against the VM resource budget, similar to how `execute()` enforces `max_instructions`.

**Tests:** Add oversized count/payload tests asserting bounded failure.

anakrish:
### Low: Runtime documentation contradicts new run-to-completion behavior

**Location:** `docs/rvm/vm-runtime.md:50-51` and `:78-79`

The documentation says preloaded responses are supported in run-to-completion mode, but also says encountering `HostAwait` is not allowed and raises an error. It also describes identifiers as function names despite explicit arbitrary IDs.

This can cause integrators to configure the wrong execution mode or API.

**Suggestion:** Update documentation to clarify:
- Run-to-completion mode preloads responses for each identifier.
- Identifiers can be explicit, arbitrary values (strings in the current FFI, but potentially more in future versions).
- Suspendable mode pauses and exposes the identifier/argument for the host to respond.

**Tests:** Documentation changes do not require tests, but the above clarifications should be verified by re-reading.

anakrish:
### Medium: Partial C# marshalling failures leak resources

**Location:** `bindings/csharp/Regorus/ModuleMarshalling.cs:316-331`, cleanup `:349-369`

Newly acquired resources are appended to tracking lists only after fallible operations. If an allocation/OOM occurs during `pins.Add`, `innerBuffers.Add`, or `innerHandles.Add`, the just-created resource is absent from cleanup and leaks:

```csharp
var valuePinned = Utf8Marshaller.Pin(kvp.Value[j]);
pins.Add(valuePinned);  // Can throw; valuePinned may not be in pins
innerBuffer[j] = (IntPtr)valuePinned.Pointer;
```

**Suggestion:** Use ownership guards or `finally` immediately on acquisition, or use rollback-safe tracking so every resource is cleaned on partial construction failure.

**Tests:** Add OOM/allocation-failure simulation to `PinHostAwaitResponseSets` and other marshalling methods.

bindings/ffi/src/rvm.rs
kusha: This landed here after all.

The stride validation added there checks base-pointer alignment as well as the stride itself, so the &* reference formation can no longer be reached with a misaligned base. 

The sibling thread on the one-directional forward-compat claim is still a follow-up, that one I haven't touched.

kusha:
Went through all eight (1 and 2 are the same one). Fixed four, pushing back on three.

**1 & 2 — FFI strides.** Done. Both entry points now go through a single `validate_array_stride` before any pointer math: size floor, stride alignment, base alignment, and total span within `isize::MAX`.

Two deliberate differences from your list. I added the base-pointer check — it wasn't in this list, though your earlier comment did mention it, and without it a misaligned base defeats an otherwise perfect stride. And I collapsed your last three bullets into the single span check, since `len * stride <= isize::MAX` already implies both `stride <= isize::MAX` (len is at least 1) and the last-element bound (stride is at least `size_of`). Same coverage, less to read — reasoning's in the doc comment if you want to check my logic.

Tests cover all four stride shapes on both structs.

**3 — Embedded NULs in response values.** Pushing back, because this reverses a call from your Aug 7 review, where you had it under "intentional and already documented, not blocking" — deliberately not validated for values, consistent with the `AddDataJson`/`SetInputJson` contract. Identifiers get checked because they're the address of a host-await with no parser behind them; values are payload and `from_json_str` catches the rest.

The module-content and data/input half is pre-existing across all the bindings, so I'd rather not quietly change that inside a host-await PR.

**4 — `SetContextJson` without `azure_policy`.** Real bug, but not relevant to FFI - fixing in a separate PR.

**5 — Non-string identifiers.** Fair, but deliberate - we landed on string-only back in June on the identifier round-trip thread. The VM handles arbitrary `Value` identifiers fine; it's the FFI that's string-only, because the identifier goes across as a raw C string so it can be handed straight back to `set_host_await_responses`. Non-string ones error cleanly rather than getting coerced into something wrong.

I've documented that properly now - the docs say the identifier is an arbitrary `Value` and that string-only is a binding limitation, not a VM one. Lifting it needs either JSON-encoded identifiers (which breaks the round-trip) or a parallel value-based API, and that's an API change I'd rather make deliberately. Enforcing it in `type_check.rs` is the other direction, but that would reject policies that run fine today.

**6 — Response queue limits.** Added, in the three accumulation loops.

Worth flagging that the gap was narrower than the finding suggests: `from_json_str` already enforces the limit inside a parse, per array element and per object entry, so big payloads were covered. What got through was lots of small ones — scalars especially, where there's nothing to iterate. And I'd call this convention-consistency rather than closing a real exhaustion vector, since the caller has to be holding the JSON already to hand it to us, and anyone in-process can exhaust memory without this API.

No test, and I'd rather say why than have it look like an oversight. The limit is process-global, and this crate has no `rlib`, so there's no separate test binary to isolate it the way `tests/memory_limits.rs` does. As a unit test it would set a global limit while ~50 others run alongside it. `ffi_roundtrips_global_limit` already has this problem — short window so it gets away with it, but a panic in there would leak the limit into the rest of the run. Filed as #804. I did check the enforcement works by hand: 1-byte limit, and it rejects.

**7 — Runtime docs.** Fixed, and you were right they can't both be true. The code settles it — `next_host_await_response` pops the queued response and only errors when there isn't one, so `HostAwait` doesn't suspend in run-to-completion, it consumes. Also fixed the identifier description, which called it a function name; that's only true for registered builtins, since the raw form passes whatever the second argument evaluates to.

**8 — Marshalling leak.** Fixed, though the cause was narrower than "partial construction failure". `pins` was sized to the identifier count but takes an entry per identifier *and* per value, so anything with more than one value forces a resize mid-loop — and that resize is the only realistic way `Add` fails, right when a pinned handle exists but isn't tracked yet. Pre-counting the values removes it. `PinModules` already does the same thing with `count * 2`.

I skipped the ownership guards: once capacity is reserved, `Add` is a bounds check and a write, so a failure there means the process is going down regardless. Say the word if you'd rather have it uniform across all four call sites. Same for the OOM test — simulating allocation failure in .NET needs heap exhaustion or a custom allocator, and both are flaky in a smoke-test suite, but I'll take a pattern if you have one.

Rebased onto current main, which needed a conflict resolution where this overlapped #792.

anakrish (approved):
LGTM

PR #695, opened 2026-04-27
feat(azure_policy): test suite, compiler bug fixes, and example program

anakrish:
This rounds out the Azure Policy compiler with a YAML-driven test suite and an example program showing the full evaluation pipeline.

## Test suite

Adds a comprehensive set of YAML test cases covering conditions, effects, template functions, alias resolution, and end-to-end scenarios adapted from real-world Azure policies. The test runner handles resource normalization, host-await responses for AINE/DINE policies, and structured effect validation.

## Compiler bug fixes

The test suite surfaced several bugs:

- **Register reuse across short-circuit boundaries:** cached parameter defaults could be stale after an allOf/anyOf arm was skipped.
- **DINE details compilation** tried to evaluate ARM deployment template fields that aren't part of policy evaluation.
- **Datetime output** wasn't consistently normalized to ISO 8601.
- **Column limit** too low for deeply nested template expressions.
- **guid/uniqueString dispatch** called unregistered builtins instead of returning a clear error.

## Example program

Restructures the example binary into an idiomatic multi-file layout and adds `azure-policy-eval` and `azure-policy-aliases` subcommands demonstrating the full pipeline: load aliases, parse a policy definition, compile, normalize an ARM resource, and evaluate in the VM.

## C# and documentation

- Adds `AzurePolicyTests.cs` exercising AliasRegistry normalization and denormalization.
- Adds an Azure Policy section to the README.


anakrish:
Splitting into two PRs to enable Copilot code review (original exceeded 20K line limit).

PR #696, opened 2026-04-27
test(azure_policy): add YAML test case files for compiler validation

anakrish:
Add 74 YAML-driven test case files covering the full spectrum of Azure Policy compiler scenarios. These data files are consumed by the test runner in the companion PR.

**Categories covered:**
- Foundation: field conditions, logical operators, type matching, exists checks
- Count/count.where expressions
- Template functions: string manipulation, datetime, numeric operations
- Advanced: nested allOf/anyOf, cross-resource effects (DINE/modify/append)
- Infrastructure patterns: DNS, networking, diagnostics
- End-to-end: 50 real-world policies from the Azure built-in library

This is a data-only PR (no code changes). The test runner that exercises these files is in the companion PR #TBD.

Part 1 of 2 — split from original #695 to enable Copilot code review on the code PR.

### DIFF
--- a/tests/azure_policy/cases/aliases.yaml
+++ b/tests/azure_policy/cases/aliases.yaml
@@ -1,0 +1,1131 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Alias Resolution Test Suite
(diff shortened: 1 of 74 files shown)


anakrish:
Splitting further into two test data PRs to enable Copilot review on the test files themselves.

PR #708, opened 2026-05-04
fix: feature-matrix CI failures for minimal and rvm test combos

anakrish:
Fixes the failures in the `tests/feature-matrix` weekly CI run ([run #25244035824](https://github.com/microsoft/regorus/actions/runs/25244035824)).

## Root Causes

**1. Missing per-feature skip logic in interpreter tests ("minimal" job)**

Tests for `time`, `semver`, `glob`, `uuid`, and `regex` builtins ran even when those features weren't enabled. The existing skip logic only handled the `no_std` case.

**2. `test_utils` module unavailable to integration tests ("library", "observability", "cherry-picked builtins" jobs)**

`test_utils` was gated behind `cfg(any(test, all(feature = "yaml", feature = "std")))`. Since `cfg(test)` doesn't apply to the library when compiled for integration tests, it required the `yaml` product feature. Integration tests with `rvm` but without `yaml` couldn't access the module.

## Fix

- Add per-feature `#[cfg(not(feature = "..."))]` skip blocks for time, semver, glob, uuid, and regex builtin tests.
- Introduce a zero-dep `test-utils` feature that exposes `test_utils` without requiring `yaml`.
- Replace `serde_yaml` usage in `test_utils::match_values` with `serde_json` (always available) so the module has no optional-dep requirements.
- Update the feature-matrix workflow to add `test-utils` to rvm-based jobs.

### DIFF
--- a/.github/workflows/feature-matrix.yml
+++ b/.github/workflows/feature-matrix.yml
@@ -30,7 +30,7 @@ jobs:
           # Common library usage pattern (issue #595): consumer enables
           # std + arc + rvm and relies on indexmap/std propagation.
           - name: library (std + arc + rvm)
-            features: std,arc,rvm
+            features: std,arc,rvm,test-utils
 
           # New default after removing mimalloc from full-opa.
           # Ensures all builtins compile without the allocator.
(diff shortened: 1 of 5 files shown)


anakrish:
@copilot review this PR using code-review skill and using deep-review skill

PR #720, opened 2026-05-11
feat(ffi): add Azure Policy JSON compilation FFI and C# bindings

anakrish:
This PR exposes Azure Policy JSON compilation through the FFI layer and C# bindings, enabling callers to compile and evaluate Azure Policy definitions natively without translating to Rego first.

### What's new

Three new FFI entry points:

- `regorus_compile_azure_policy_rule` and `regorus_compile_azure_policy_definition` parse and compile JSON policy rules/definitions into RVM programs, with optional alias resolution via the existing `RegorusAliasRegistry`.
- `regorus_rvm_set_context` lets callers supply ambient host data (`subscription()`, `resourceGroup()`, etc.) that context-dependent policies need at evaluation time.

On the C# side, there's a new `AzurePolicyCompiler` helper class and a `Rvm.SetContextJson` method, along with tests and README docs.

### Notable choices

- `set_context` validates that the JSON is an object (returns `InvalidDataFormat` otherwise) and persists across executions, matching the existing `set_input` behavior.
- The compile functions accept a null registry for simple policies that don't use aliases — this is documented with clear warnings about the implications.
- Uses `Rc::try_unwrap` to avoid cloning the compiled program when the refcount is 1 (the common case).
- Everything is behind `#[cfg(feature = "azure_policy")]`.
- FFI crate version bumped to 0.10.1.

### DIFF
--- a/bindings/csharp/Directory.Packages.props
+++ b/bindings/csharp/Directory.Packages.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-    <RegorusPackageVersion>0.10.0</RegorusPackageVersion>
+    <RegorusPackageVersion>0.10.1</RegorusPackageVersion>
     <RegorusPackageVersionSuffix Condition="'$(VersionSuffix)' != ''">-$(VersionSuffix)</RegorusPackageVersionSuffix>
   </PropertyGroup>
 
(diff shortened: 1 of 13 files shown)


anakrish:
@copilot Use the deep-review skill to review this PR.

anakrish:
Superseded by new PR with cleaned up API surface (removed CompilePolicyRule, only CompilePolicyDefinition exposed).

PR #737, opened 2026-06-02
chore(rvm): add debug-mode invariant assertions

anakrish:
Add `debug_assert!`-based invariant checks to the RVM covering:

- `reset_execution_state` postcondition (all stacks empty, registers reset, caches cleared, state `Ready`).
- Per-opcode invariants in the dispatch loop (state, registers, rule cache, execution-stack sanity bound).
- `resume` and `execute_suspendable_entry` preconditions.
- Push/pop balance on the comprehension and call-rule stacks.
- `IterationState::advance` (no double-consumption, no index overflow).

All assertions are `#[cfg(debug_assertions)]`-gated, so release builds are unaffected. No existing tests required modification.

### DIFF
--- a/src/rvm/vm/comprehension.rs
+++ b/src/rvm/vm/comprehension.rs
@@ -528,7 +528,6 @@ impl RegoVM {
             Ok(false)
         }
     }
-
     pub(super) fn handle_comprehension_condition_failure_suspendable(&mut self) -> Result<bool> {
         if let Some(mut frame) = self.execution_stack.pop() {
             let handled = if let &mut FrameKind::Comprehension {
@@ -606,6 +605,9 @@ impl RegoVM {
     }
 
     fn execute_comprehension_end_run_to_completion(&mut self) -> Result<()> {
+        // `ComprehensionEnd` is reached from a loaded program; an empty stack
+        // here means malformed user-supplied bytecode, which must still surface
+        // as a typed error rather than a panic — including in debug builds.
         self.comprehension_stack.pop().map_or_else(
             || {
                 Err(VmError::InvalidIteration {
--- a/src/rvm/vm/context.rs
+++ b/src/rvm/vm/context.rs
@@ -54,6 +54,14 @@ impl IterationState {
     pub(super) const fn advance(&mut self) {
         match *self {
             Self::Array { ref mut index, .. } => {
+                // Array iteration uses `usize` as the cursor and advances via
+                // `saturating_add(1)`. A cursor already at `usize::MAX` here
+                // means a stuck (non-progressing) iteration was emitted by
+                // malformed bytecode; assert in debug to surface it loudly.
+                debug_assert!(
+                    *index < usize::MAX,
+                    "IterationState::Array index already at usize::MAX on advance"
+                );
                 *index = index.saturating_add(1);
             }
             Self::Object {
@@ -69,6 +77,12 @@ impl IterationState {
             Self::Single {
                 ref mut consumed, ..
             } => {
+                // `Single` yields exactly once; advancing a consumed Single
+                // means the compiler emitted a redundant LoopNext.
+                debug_assert!(
+                    !*consumed,
+                    "IterationState::Single advanced after consumption"
+                );
                 *consumed = true;
             }
         }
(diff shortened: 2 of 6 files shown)


anakrish (commented):
Thanks — both findings are valid and addressed in the latest force-push (head `1662571`):

1. **`resume()` debug_assert** (`execution.rs:212-216`): removed. You're right that it diverged debug vs release behavior on a recoverable misuse, and the FFI-poisoning consideration is the deciding factor — `resume()` is exactly the kind of API a host can call out of order without that being an internal VM bug. The existing `VmError::InvalidResumeState` return remains the sole enforcement; left a comment explaining why no debug assertion belongs here.

2. **`IterationState::Array` comment** (`context.rs:57-63`): the comment was inaccurate — `saturating_add` saturates rather than wraps. Rewrote the comment to describe what the assertion actually catches: an array cursor already at `usize::MAX` on advance, which would mean a non-progressing (stuck) iteration. Updated the panic message to match.

Validation: `cargo +1.92.0 xtask fmt`, `xtask clippy --all-targets --all-features -- -D warnings`, and `cargo +1.92.0 test --lib --all-features` (578 passed) all green.

anakrish (commented):
Thanks — both findings are valid and addressed in the latest force-push (head `fb31788`):

1. **`state.rs:11-14` doc comment**: corrected. The RVM's actual execution-limit mechanisms are `RegoVM::set_max_instructions` (checked in the dispatch loop) and `memory_check` — `enforce_limit` exists but is the builtin-recursion limiter in `src/builtins/utils.rs`, not an RVM limit; `MachineLimits` doesn't exist. Doc now references the right names.

2. **`rules.rs:283-287` debug_assert_eq! on rule_index mismatch**: converted to a typed `VmError::CallRuleStackMismatch { expected, actual, pc }` returned in all builds. You're right that a debug-only panic on this condition was the wrong call — it can be triggered by malformed/poisoned program state (same class as the underflow case immediately above), and per the Copilot-instructions invariant "Panics in FFI = permanent poisoning" any path reachable from a host call must surface a recoverable error rather than a panic. Added the variant to `VmError` and updated the call site with a comment explaining the discipline.

Validation: `cargo +1.92.0 xtask fmt`, `xtask clippy --all-targets --all-features -- -D warnings`, `cargo +1.92.0 test --lib --all-features` (578 passed) all green.

anakrish:
@copilot review this PR using each review skill in this repo. Use a separate agent for each skill.

anakrish (commented):
Self-review pass on top of the previous bot rounds — applying the same "any debug_assert reachable from host-controllable input must be a typed VmError" discipline systematically across the diff rather than per-flagged-site. Pushed as `4c42093`.

Findings and fixes:

1. **`assert_vm_invariants`: `!self.registers.is_empty()` was host-reachable.** `RuleInfo::num_registers: u8` is plain serde-derived with no validator (`src/rvm/program/types.rs:96`). A loaded program with `num_registers == 0` for a `RuleType::Complete` rule called outside a function would, after `execute_rule_definitions_common`'s `self.registers.resize(num_registers, …)` (`rules.rs:103-105`), leave registers empty — and the next dispatch iteration would panic in debug. Same FFI-poisoning class as the two earlier findings. Removed the assertion; the existing `VmError::RegisterIndexOutOfBounds` already catches the real failure at the first register access.

2. **`assert_vm_invariants`: `execution_stack.len() <= 4096` was host-reachable.** A malformed program with deeply nested rules/loops/comprehensions can push past 4096 frames in suspendable mode well within the default `max_instructions` budget, triggering the debug panic. Removed the assertion (and the `DEBUG_MAX_EXECUTION_STACK_DEPTH` constant); runaway recursion is bounded in production by `set_max_instructions` and `memory_check`, which is where this concern actually belongs.

3. **Kept** the two genuinely VM-internal invariants in `assert_vm_invariants`: `execution_state ∈ {Ready, Running}` (VM-internal state machine) and `rule_cache.len() == program.rule_infos.len()` (sized at our own reset; never mutated mid-execution). Also kept `debug_assert_state_is_clean` postconditions as-is — those are postconditions on our own reset code, not reachable from host input.

The remainder of the audit (other `debug_assert!`s in the diff, doc-comment symbol fabrications, comment-vs-code drift, unchecked arithmetic, stale invariants) came back clean.

Validation: `cargo +1.92.0 xtask fmt`, `xtask clippy --all-targets --all-features -- -D warnings`, `cargo +1.92.0 test --lib --all-features` (578 passed) all green.

dpokluda (approved):

Issue #743, opened 2026-06-06
Behaviour differing from OPA on cyclic references between multiple policies


phi1010:
The results from OPA and regorus differ for noncyclic rules across cyclicly referencing packages/files.

I get 

```jsonl
opa eval --data test/registry.rego --data test/package_a.rego data
Allowed by stage 1:  package_a
Allowed by stage 2:  package_a
{
  "result": [
    {
      "expressions": [
        {
          "value": {
            "registry": {
              "allow": true,
              "allow_stage1": true,
              "allow_stage2": true,
              "packages": {
                "package_a": {
                  "allow_stage1": true,
                  "allow_stage2": true
                }
              }
            }
          },
          "text": "data",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

with OPA 1.17.0 while regorus v0.9.1 with python using the `Engine` class returns:

```
/home/phi1010/.local/bin/uv run /home/phi1010/PycharmProjects/htmlplayground1/.venv/bin/python /home/phi1010/PycharmProjects/htmlplayground1/documentation/evaluate_policies_with_regorus.py 
ic| engine: <builtins.Engine object at 0xb30ca00>
Traceback (most recent call last):
  File "/home/phi1010/PycharmProjects/htmlplayground1/documentation/evaluate_policies_with_regorus.py", line 22, in <module>
    run()
    ~~~^^
  File "/home/phi1010/PycharmProjects/htmlplayground1/documentation/evaluate_policies_with_regorus.py", line 18, in run
    ic(engine.eval_query("data"))
       ~~~~~~~~~~~~~~~~~^^^^^^^^
RuntimeError: 
--> package_a.rego:6:1
  |
6 | allow_stage2 if registry.allow_stage1
  | ^
error: recursion detected when evaluating rule:
--> package_a.rego:6:1
  |
6 | allow_stage2 if registry.allow_stage1
  | ^
depends on: 
--> registry.rego:5:1
  |
5 | allow if allow_stage2
  | ^
depends on: 
--> registry.rego:12:1
   |
12 | allow_stage2 if {
   | ^
depends on: 
--> package_a.rego:6:1
  |
6 | allow_stage2 if registry.allow_stage1
  | ^
depends on: cyclic evaluation

Process finished with exit code 1
```

The files look like this:

package_a.rego
```rego
package registry.packages.package_a

import data.registry
import rego.v1

allow_stage2 if registry.allow_stage1

[...]

PR #762, opened 2026-07-13
Bugfix/partially cyclic evaluation

phi1010:
Bugfix written with Fable5 alone, and unit tests for the edgecases it fixes.
Code not yet verified by a Rust developer, might cause unintended side effects the tests don't cover.
Please review thouroughly as discussed in #743

### DIFF
--- a/docs/cross-package-lazy-evaluation.md
+++ b/docs/cross-package-lazy-evaluation.md
@@ -154,9 +154,61 @@ matching OPA's verdict, while the valid issue #743 policies re-evaluate
 to identical values and pass. The check only runs when a partial sweep
 actually occurred, so well-behaved policies pay no cost.
 
-Regression tests: `cross-module-package-cycle` and
-`cross-module-package-cycle-vanishing` in
-`tests/interpreter/cases/rule/dependency.yaml`.
+### Deferred back-references through a sweep
+
+A swept rule may itself directly reference the rule that initiated the
+sweep. With only one rule in the registry package:
+
+```rego
+package registry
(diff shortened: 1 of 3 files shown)


phi1010:

phi1010:
This is still incomplete, it does not reject all recursions in doubt:
```py
import textwrap

import regorus
from icecream import ic


def run():

    eng = regorus.Engine()
    eng.add_policy(
        "recursion.rego",
        textwrap.dedent(
            """
            package foo
            import rego.v1
            a := [input.x,i] if {
                i:=a with input as { "x": input.x+1 }
            }    
            """),
    )
    ic(eng.eval_query("data"))

if __name__ == "__main__":
    run()

```

```sh
$ ./rego-recursion.py 
ic| eng.eval_query("data"): {'result': [{'bindings': {},
                                         'expressions': [{'location': {'col': 1, 'row': 1},
                                                          'text': 'data',
                                                          'value': {'foo': {}}}]}]}


$ opa eval --data test.rego  "data"
{
  "errors": [
    {
      "message": "rule data.foo.a is recursive: data.foo.a -\u003e data.foo.a",
      "code": "rego_recursion_error",
      "location": {
        "file": "test.rego",
        "row": 3,
        "col": 13
      }
    }
  ]
}
```

anakrish:
@phi1010 Can you also address the RVM flow in this PR? Currently the fix only targets the interpreter.

I will free up next week and should be able to devote more time to this.

phi1010:
I found an easier solution for my usecase, i just wrote some bindings to the go library: https://pypi.org/project/opa-golib-python-bindings/

PR #769, opened 2026-07-23
fix(interpreter,rvm): resolve function calls through import aliases

### LINKED ISSUE #747: Function calls through imported packages fail with "could not find function"
Hi, it seems like functions cannot be called through package imports — only variable refs resolve through imports (related: #541, #284), see the example below:

```rego
package lib

import rego.v1

double(x) := 2 * x
```

```rego
package rules

import data.lib
import rego.v1

result := lib.double(21)
```

Evaluating `data.rules.result` after loading these two policies fails with:

```
Error:
--> rules.rego:6:21
  |
6 | result := lib.double(21)
  |                     ^
error: could not find function lib.double
```

Using an explicit alias (`import data.lib as l` / `l.double(21)`) fails the same way. Referring to the function by its fully qualified name (`data.lib.double(21)`) works, and so do variable refs through the same import (e.g. `lib.some_value`). OPA evaluates the sample above to `42`.

The compiled path is affected as well: `Compiler::compile_from_policy` rejects the same program at compile time with `error: Unknown function: 'lib.double'` for any entrypoint that reaches the call.

I have tested it against commit ed6ae465b0668b9602d2589875d29a65e55bf4de (and it reproduces identically on the v0.10.1 tag).

I'm happy to send a PR for both paths if that helps.

michaelstaib:
This PR fixes function-call resolution through package imports (#747). Calls like `helpers.is_admin(x)` after `import data.lib.helpers` failed with `could not find function` in both evaluation paths, while value references through the same imports worked. Both engines now resolve the call path's leading component through the calling module's imports, mirroring the existing value-reference fallbacks and matching OPA semantics: the plain, `as`-alias, and bare member-import forms all resolve, and an import alias takes precedence over builtin namespaces and module-local ref-head rules of the same spelling.

### Key Components

- **Interpreter Call Resolution**: `eval_call_impl` rewrites a leading import alias before the `with`-override and builtin lookups, so overrides keyed by full paths reach aliased calls and an alias shadows a like-named builtin namespace.

- **RVM Compiler Call Resolution**: `compile_function_call` consults the calling module's imports before module-prefixing, with the same precedence as the interpreter.

- **Resolution Gating**: A path is only rewritten when it resolves to a known function, or in the interpreter also an exact `default` function (`rule_paths` membership plus a non-empty argument list). The `rules`/`default_rules` maps cannot serve as the gate because they also index value rules and every rule-path prefix, none of which may become callable through an alias. `rego.v1`/`future.keywords` imports stay inert, unknown functions error cleanly, and modules without imports skip resolution entirely. When the aliased package does not define the called member, the call keeps its prior meaning (e.g. a builtin call) instead of erroring like OPA; the gate is deliberate so the rewrite never changes the meaning of existing policies.

- **`with` Modifier Targets**: A leading alias in a `with` target is rewritten unconditionally, matching OPA: a target resolving to a function registers a function override, and anything else becomes a data or input override (`with fl.flag as 5` behaves exactly like `with data.flib.flag as 5`). The rewrite splices parsed reference components rather than round-tripping through a dot-joined string, so bracketed keys containing dots survive on both sides of the alias. Interpreter only, since the RVM compiler rejects the keyword (`WithKeywordUnsupported`).

- **Module Restore**: `eval_call_impl` now restores the caller's module when a function overload errors. The callee's module previously leaked, which would make the rest of the caller's body resolve aliases through the wrong module's imports when the error is swallowed in non-strict mode.

- **Tests**: 23 new interpreter cases and 13 new RVM cases in the import suites, covering all call forms, precedence, gating rejections, `with` interactions (including dotted bracket keys), and regressions. 
- 
- Expected outputs are cross-checked against **OPA v1.18.2**. A pre-existing RVM limitation (cross-package calls to `default`-valued functions fail with "not a valid rule path") is skip-marked with a TODO and covered in the interpreter suite.

### Testing

`cargo test --locked`, `cargo fmt --check`, and `cargo clippy --all-targets --all-features -- -D warnings` (plus `--no-default-features` builds and the `ensure_no_std` target check) all pass.

Fixes #747. Builds on @andrioni's unsubmitted fix from the issue thread (andrioni/regorus@ff2f8bee5d59440ecfc97d48fef31b8017bec867).

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -1258,7 +1258,28 @@ impl Interpreter {
             // Apply with modifiers.
             for wm in &stmt.with_mods {
                 let path = Parser::get_path_ref_components(&wm.refr)?;
-                let path: Vec<&str> = path.iter().map(|s| s.text()).collect();
+                let mut path: Vec<String> = path.iter().map(|s| s.text().to_string()).collect();
+
+                // Matching OPA, a leading import alias is rewritten before
+                // any lookups: functions register as overrides below,
+                // anything else becomes a data override. Only the alias
+                // component is replaced so bracketed keys containing dots
+                // survive the rewrite.
+                let rewritten: Option<Vec<String>> = match path.split_first() {
+                    Some((head, rest)) if head.as_str() != "data" => {
+                        self.lookup_import_alias(head).map(|import_target| {
+                            import_target
+                                .split('.')
+                                .map(str::to_string)
+                                .chain(rest.iter().cloned())
+                                .collect()
+                        })
+                    }
+                    _ => None,
+                };
+                if let Some(new_path) = rewritten {
+                    path = new_path;
+                }
                 let mut target = path.join(".");
 
                 let mut target_is_function = self.lookup_function_by_name(&target).is_some()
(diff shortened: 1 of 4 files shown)


anakrish (approved):
@michaelstaib Thanks for your contribution. Looks good!

I've captured some corner cases/follow up items in https://github.com/microsoft/regorus/issues/770

PR #792, opened 2026-08-19
Add per-execution memory budgets to RVM

maksym-mishchenko:
AB#3522638

A process-global allocator limit cannot isolate individual policy evaluations and may cause unrelated requests to fail. This adds optional per-execution memory budgets for run-to-completion RVM evaluations.

Enforcement is cooperative and checkpoint-based. The budget observes current-thread live bytes relative to a fresh execution baseline; it is not an allocation-time peak cap, so one instruction, builtin, native serialization step, or `CString` allocation may temporarily overshoot. Reported usage is a diagnostic thread-level change rather than exact query-owned memory. Synchronous callbacks on the execution thread affect accounting, cross-thread frees may temporarily skew observations, and a lower live-byte sample ratchets the baseline downward without restoring lost headroom.

A reused VM captures a fresh baseline after prior execution state is released, but retained capacities and pools precede that baseline, so warm and fresh VMs may allocate differently. Program compilation and loading data, input, or context remain outside the execution budget. Native result serialization and `CString` construction are included; managed C# UTF-8 decoding and managed-string allocation are excluded.

Budget failures return additive typed Rust, FFI, and C# errors, invalidate completed results, release retained execution state, and take precedence over the process-global allocator limit. Budgeted suspendable execution is rejected because thread-local accounting cannot safely span host-await thread migration; host-await accounting is separate follow-up work.

The opt-in controls are `RegoVM::set_memory_budget_config` and C# `Rvm.SetMemoryBudgetConfig`; existing behavior is unchanged when no budget is configured.

### DIFF
--- a/benches/rvm_benchmark.rs
+++ b/benches/rvm_benchmark.rs
@@ -38,6 +38,8 @@
 
 use std::hint::black_box;
 use std::num::NonZeroU32;
+#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+use std::num::NonZeroU64;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
@@ -50,6 +52,8 @@ use regorus::languages::rego::compiler::Compiler;
 use regorus::rvm::program::Program;
 use regorus::rvm::vm::{ExecutionMode, RegoVM};
 use regorus::utils::limits::ExecutionTimerConfig;
+#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+use regorus::MemoryBudgetConfig;
 use regorus::{Engine, Rc, Value};
 
 // ---------------------------------------------------------------------------
@@ -57,7 +61,7 @@ use regorus::{Engine, Rc, Value};
 // hot path (memory_check, execution_timer_tick, instruction-limit compare).
 // ---------------------------------------------------------------------------
 
-#[cfg(feature = "allocator-memory-limits")]
+#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
 const MEMORY_LIMIT_BYTES: u64 = 256 * 1024 * 1024;
 const TIME_LIMIT: Duration = Duration::from_secs(30);
 const TIMER_CHECK_INTERVAL: NonZeroU32 = NonZeroU32::new(16).unwrap();
@@ -68,28 +72,39 @@ struct EvalConfig {
     name: &'static str,
     mode: ExecutionMode,
     limits: bool,
+    memory_budget: bool,
 }
 
-const EVAL_CONFIGS: [EvalConfig; 4] = [
+const EVAL_CONFIGS: [EvalConfig; 5] = [
     EvalConfig {
         name: "regular_no_limits",
         mode: ExecutionMode::RunToCompletion,
         limits: false,
+        memory_budget: false,
+    },
+    EvalConfig {
+        name: "regular_memory_budget",
+        mode: ExecutionMode::RunToCompletion,
+        limits: false,
+        memory_budget: true,
     },
     EvalConfig {
         name: "regular_with_limits",
         mode: ExecutionMode::RunToCompletion,
         limits: true,
+        memory_budget: true,
     },
     EvalConfig {
         name: "suspendable_no_limits",
         mode: ExecutionMode::Suspendable,
         limits: false,
+        memory_budget: false,
     },
     EvalConfig {
         name: "suspendable_with_limits",
         mode: ExecutionMode::Suspendable,
         limits: true,
+        memory_budget: false,
     },
 ];
 
@@ -358,10 +373,13 @@ fn compile_all_programs() -> Vec<BenchmarkProgram> {
 // Limit helpers
 // ---------------------------------------------------------------------------
 
-/// Apply or remove production-style limits based on a boolean flag.
-fn configure_limits(vm: &mut RegoVM, limits: bool) {
-    if limits {
-        #[cfg(feature = "allocator-memory-limits")]
+/// Apply the limits selected for one benchmark configuration.
+fn configure_limits(vm: &mut RegoVM, config: EvalConfig) {
+    #[cfg(any(miri, not(feature = "allocator-memory-limits")))]
+    let _ = config.memory_budget;
+
+    if config.limits {
+        #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
         regorus::set_global_memory_limit(Some(MEMORY_LIMIT_BYTES));
         vm.set_execution_timer_config(Some(ExecutionTimerConfig {
             limit: TIME_LIMIT,
@@ -369,11 +387,16 @@ fn configure_limits(vm: &mut RegoVM, limits: bool) {
         }));
         vm.set_max_instructions(INSTRUCTION_LIMIT);
     } else {
-        #[cfg(feature = "allocator-memory-limits")]
+        #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
         regorus::set_global_memory_limit(None);
         vm.set_execution_timer_config(None);
         vm.set_max_instructions(usize::MAX);
     }
+
+    #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+    vm.set_memory_budget_config(config.memory_budget.then(|| MemoryBudgetConfig {
+        limit: NonZeroU64::new(MEMORY_LIMIT_BYTES).expect("non-zero memory budget"),
+    }));
 }
 
 // ---------------------------------------------------------------------------
@@ -408,7 +431,7 @@ fn bench_cold(c: &mut Criterion) {
                             vm.set_data(black_box(d.clone())).unwrap();
                         }
                         vm.set_input(black_box(input.clone()));
-                        configure_limits(&mut vm, config.limits);
+                        configure_limits(&mut vm, config);
                         black_box(vm.execute().unwrap())
                     })
                 });
@@ -445,7 +468,7 @@ fn bench_hot(c: &mut Criterion) {
                 if let Some(ref d) = data {
                     vm.set_data(d.clone()).unwrap();
                 }
-                configure_limits(&mut vm, config.limits);
+                configure_limits(&mut vm, config);
 
                 // Warm up: fill register window pools, caches, etc.
                 vm.set_input(inputs[0].clone());
--- a/bindings/csharp/README.md
+++ b/bindings/csharp/README.md
@@ -105,6 +105,35 @@ var result = vm.Execute();
 Console.WriteLine($"allow: {result}");
 ```
 
+### Per-execution memory budget
+
+RVM run-to-completion evaluation can use an optional additional live-memory budget. Each ordinary `Execute` or `ExecuteEntryPoint` call starts with a fresh budget for execution; program compilation, program loading, and prior `SetDataJson`, `SetInputJson`, and `SetContextJson` calls occur before and outside that budget.
+
+```csharp
+using var vm = new Rvm();
+vm.LoadProgram(program);
+vm.SetDataJson(Data);
+vm.SetInputJson(Input);
+vm.SetMemoryBudgetConfig(new MemoryBudgetConfig(16 * 1024 * 1024));
+
+try
+{
+    var result = vm.Execute();
+}
+catch (RegorusMemoryBudgetExceededException ex)
+{
+    Console.WriteLine(ex.Message);
+}
+```
+
+The native execution budget includes result JSON serialization and Rust `CString` allocation before the native call returns. Managed UTF-8 decoding and C# `string` allocation after that return are not charged.
+
+This is a cooperative observed-live-bytes budget, not an allocation-time peak-memory cap. A single instruction, builtin, native serialization, or `CString` allocation may exceed the limit before the next checkpoint; configure headroom for that overshoot.
+
+Accounting observes the execution thread rather than allocation ownership. Synchronous host work on that thread contributes to usage, cross-thread frees can temporarily skew observations, and a downward baseline ratchet can permanently remove headroom during an execution. A reused VM receives a new baseline, but retained capacities and pools can make it allocate differently from a fresh VM. See the [RVM memory budget documentation](../../docs/limits/memory_budget.md) for the detailed accounting model.
+
+Budgets are not supported in suspendable execution mode. `ClearMemoryBudgetConfig` restores the previous unlimited per-execution behavior. Public multi-call begin/end scopes are intentionally absent because allocator counters are thread-local. Failed terminal execution clears retained state. The process-wide limit exposed by `MemoryLimits` remains a separate safeguard.
+
 ## Azure RBAC Condition Evaluation
 
 Evaluate Azure RBAC condition expressions directly with a JSON evaluation context:
--- a/bindings/csharp/Regorus/MemoryBudgetConfig.cs
+++ b/bindings/csharp/Regorus/MemoryBudgetConfig.cs
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Regorus
+{
+    /// <summary>
+    /// Configures the additional live-memory budget for one RVM execution.
+    /// </summary>
+    public readonly struct MemoryBudgetConfig
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MemoryBudgetConfig"/> struct.
+        /// </summary>
+        /// <param name="limitBytes">Maximum additional live bytes allowed during one execution.</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="limitBytes"/> is zero.</exception>
+        public MemoryBudgetConfig(ulong limitBytes)
+        {
+            if (limitBytes == 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(limitBytes), "Memory budget must be non-zero.");
+            }
+
+            LimitBytes = limitBytes;
+        }
+
+        /// <summary>
+        /// Gets the maximum additional live bytes allowed during one execution.
+        /// </summary>
+        public ulong LimitBytes { get; }
+
+        internal Regorus.Internal.RegorusMemoryBudgetConfig ToNative()
+        {
+            if (LimitBytes == 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(LimitBytes), "Memory budget must be non-zero.");
+            }
+
+            return new Regorus.Internal.RegorusMemoryBudgetConfig
+            {
+                limit_bytes = LimitBytes,
+            };
+        }
+    }
+}
--- a/bindings/csharp/Regorus/NativeMethods.cs
+++ b/bindings/csharp/Regorus/NativeMethods.cs
@@ -245,6 +245,14 @@ namespace Regorus.Internal
         /// </summary>
         [DllImport(LibraryName, EntryPoint = "regorus_rvm_set_execution_timer_config", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
         internal static extern RegorusResult regorus_rvm_set_execution_timer_config(RegorusRvm* vm, [MarshalAs(UnmanagedType.I1)] bool has_config, RegorusExecutionTimerConfig config);
+
+        /// <summary>
+        /// Set memory budget configuration.
+        /// </summary>
+        [DllImport(LibraryName, EntryPoint = "regorus_rvm_set_memory_budget_config", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
+        internal static extern RegorusResult regorus_rvm_set_memory_budget_config(RegorusRvm* vm, [MarshalAs(UnmanagedType.I1)] bool has_config, RegorusMemoryBudgetConfig config);
+
+        /// <summary>
         /// Add a policy.
         /// The policy is parsed into AST.
         /// See https://docs.rs/regorus/latest/regorus/struct.Engine.html#method.add_policy
@@ -828,6 +836,14 @@ namespace Regorus.Internal
         /// The engine remains poisoned because a previous panic was detected.
         /// </summary>
         Poisoned,
+        /// <summary>
+        /// An RVM execution exceeded its configured memory budget.
+        /// </summary>
+        MemoryBudgetExceeded,
+        /// <summary>
+        /// An RVM memory budget was used with suspendable execution.
+        /// </summary>
+        MemoryBudgetUnsupportedInSuspendableExecution,
     }
 
     /// <summary>
@@ -881,6 +897,15 @@ namespace Regorus.Internal
     {
         public ulong limit_ns;
         public uint check_interval;
+    }
+
+    /// <summary>
+    /// FFI representation of the RVM memory budget configuration.
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct RegorusMemoryBudgetConfig
+    {
+        public ulong limit_bytes;
     }
 
     /// <summary>
--- a/bindings/ffi/src/rvm.rs
+++ b/bindings/ffi/src/rvm.rs
@@ -7,7 +7,7 @@ use crate::common::{
 };
 use crate::compile::RegorusPolicyModule;
 use crate::compiled_policy::RegorusCompiledPolicy;
-use crate::limits::RegorusExecutionTimerConfig;
+use crate::limits::{RegorusExecutionTimerConfig, RegorusMemoryBudgetConfig};
 use crate::lock::{new_handle, try_read, try_write, Handle, ReadGuard, WriteGuard};
 use crate::panic_guard::with_unwind_guard;
 use alloc::boxed::Box;
@@ -23,7 +23,7 @@ use regorus::rvm::program::{
     generate_assembly_listing, generate_tabular_assembly_listing, AssemblyListingConfig,
     DeserializationResult, Program,
 };
-use regorus::rvm::vm::{ExecutionMode, ExecutionState, RegoVM};
+use regorus::rvm::vm::{ExecutionMode, ExecutionState, RegoVM, VmError};
 use regorus::PolicyModule;
 use regorus::Value;
 
@@ -53,6 +53,53 @@ impl RegorusRvm {
 
     fn try_read(&self) -> Result<ReadGuard<'_, RegoVM>> {
         try_read(&self.vm).ok_or_else(Self::contention_error)
+    }
+}
+
+fn to_rvm_string_result(output: Result<String>) -> RegorusResult {
+    match output {
+        Ok(json) => RegorusResult::ok_string(json),
+        Err(err) => to_rvm_error_result(err),
+    }
+}
+
+fn to_rvm_error_result(err: anyhow::Error) -> RegorusResult {
+    let status = match err.downcast_ref::<VmError>() {
+        Some(VmError::MemoryBudgetExceeded { .. }) => RegorusStatus::MemoryBudgetExceeded,
+        Some(VmError::MemoryBudgetUnsupportedInSuspendableExecution { .. }) => {
+            RegorusStatus::MemoryBudgetUnsupportedInSuspendableExecution
+        }
+        _ => RegorusStatus::Error,
+    };
+    RegorusResult::err_with_message(status, err.to_string())
+}
+
+enum RvmExecution {
+    Main,
+    Named(String),
+    Indexed(usize),
+}
+
+fn execute_to_rvm_result(vm: *mut RegorusRvm, execution: RvmExecution) -> RegorusResult {
+    let output = || -> Result<RegorusResult> {
+        let vm = to_shared_ref(vm as *const RegorusRvm)?;
+        let mut guard = vm.try_write()?;
+
+        let output = match execution {
+            RvmExecution::Main => guard.execute_to_c_string_for_ffi()?,
+            RvmExecution::Named(entry_point) => {
+                guard.execute_entry_point_by_name_to_c_string_for_ffi(&entry_point)?
+            }
+            RvmExecution::Indexed(index) => {
+                guard.execute_entry_point_by_index_to_c_string_for_ffi(index)?
+            }
+        };
+        Ok(RegorusResult::ok_c_string(output))
+    }();
+
+    match output {
+        Ok(result) => result,
+        Err(err) => to_rvm_error_result(err),
     }
 }
 
@@ -478,22 +525,57 @@ pub extern "C" fn regorus_rvm_set_execution_timer_config(
     })
 }
 
+/// Configure the per-VM memory budget for run-to-completion execution.
+#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+#[no_mangle]
+pub extern "C" fn regorus_rvm_set_memory_budget_config(
+    vm: *mut RegorusRvm,
+    has_config: bool,
+    config: RegorusMemoryBudgetConfig,
+) -> RegorusResult {
+    with_unwind_guard(|| {
+        let config = if has_config {
+            match config.to_memory_budget_config() {
+                Ok(config) => Some(config),
+                Err(err) => {
+                    return RegorusResult::err_with_message(
+                        RegorusStatus::InvalidArgument,
+                        err.to_string(),
+                    )
+                }
+            }
+        } else {
+            None
+        };
+
+        to_regorus_result(|| -> Result<()> {
+            let vm = to_shared_ref(vm as *const RegorusRvm)?;
+            let mut guard = vm.try_write()?;
+            guard.set_memory_budget_config(config);
+            Ok(())
+        }())
+    })
+}
+
+/// Report that memory budgets are unavailable without allocator tracking.
+#[cfg(any(not(feature = "allocator-memory-limits"), miri))]
+#[no_mangle]
+pub extern "C" fn regorus_rvm_set_memory_budget_config(
+    _vm: *mut RegorusRvm,
+    _has_config: bool,
+    _config: RegorusMemoryBudgetConfig,
+) -> RegorusResult {
+    RegorusResult::err_with_message(
+        RegorusStatus::InvalidArgument,
+        "regorus_rvm_set_memory_budget_config unavailable: allocator memory tracking is disabled"
+            .into(),
+    )
+}
+
 /// Execute the program's main entry point.
 #[no_mangle]
 pub extern "C" fn regorus_rvm_execute(vm: *mut RegorusRvm) -> RegorusResult {
-    with_unwind_guard(|| {
-        let output = || -> Result<String> {
-            let vm = to_shared_ref(vm as *const RegorusRvm)?;
-            let mut guard = vm.try_write()?;
-            let result = guard.execute()?;
-            result.to_json_str()
-        }();
-
-        match output {
-            Ok(json) => RegorusResult::ok_string(json),
-            Err(err) => RegorusResult::err_with_message(RegorusStatus::Error, err.to_string()),
-        }
-    })
+    with_unwind_guard(|| execute_to_rvm_result(vm, RvmExecution::Main))
 }
 
 /// Execute a named entry point.
@@ -503,18 +585,11 @@ pub extern "C" fn regorus_rvm_execute_entry_point_by_name(
     entry_point: *const c_char,
 ) -> RegorusResult {
     with_unwind_guard(|| {
-        let output = || -> Result<String> {
-            let vm = to_shared_ref(vm as *const RegorusRvm)?;
-            let mut guard = vm.try_write()?;
-            let name = from_c_str(entry_point)?;
-            let result = guard.execute_entry_point_by_name(&name)?;
-            result.to_json_str()
-        }();
-
-        match output {
-            Ok(json) => RegorusResult::ok_string(json),
-            Err(err) => RegorusResult::err_with_message(RegorusStatus::Error, err.to_string()),
-        }
+        let entry_point = match from_c_str(entry_point) {
+            Ok(entry_point) => entry_point,
+            Err(err) => return to_rvm_error_result(err),
+        };
+        execute_to_rvm_result(vm, RvmExecution::Named(entry_point))
     })
 }
 
@@ -524,19 +599,7 @@ pub extern "C" fn regorus_rvm_execute_entry_point_by_index(
     vm: *mut RegorusRvm,
     index: usize,
 ) -> RegorusResult {
-    with_unwind_guard(|| {
-        let output = || -> Result<String> {
-            let vm = to_shared_ref(vm as *const RegorusRvm)?;
-            let mut guard = vm.try_write()?;
-            let result = guard.execute_entry_point_by_index(index)?;
-            result.to_json_str()
-        }();
-
-        match output {
-            Ok(json) => RegorusResult::ok_string(json),
-            Err(err) => RegorusResult::err_with_message(RegorusStatus::Error, err.to_string()),
-        }
-    })
+    with_unwind_guard(|| execute_to_rvm_result(vm, RvmExecution::Indexed(index)))
 }
 
 /// Resume execution for suspendable runs.
@@ -559,10 +622,7 @@ pub extern "C" fn regorus_rvm_resume(
             result.to_json_str()
         }();
 
-        match output {
-            Ok(json) => RegorusResult::ok_string(json),
-            Err(err) => RegorusResult::err_with_message(RegorusStatus::Error, err.to_string()),
-        }
+        to_rvm_string_result(output)
     })
 }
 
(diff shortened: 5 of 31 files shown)


src/rvm/vm/execution.rs
```diff
@@ -16,13 +16,15 @@ use super::machine::RegoVM;
 
 impl RegoVM {
     pub fn execute(&mut self) -> Result<Value> {
+        self.ensure_memory_budget_execution_mode()?;
```
anakrish, line 19: **[High] `resume()` bypasses this guard, enforcing the budget on the suspendable path with a stale/foreign baseline**

*Finding from the correctness/memory-accounting review; independently reproduced by the testing/adversarial review and the repo `code-review` skill.*

`ensure_memory_budget_execution_mode()` is called at all three `execute*` entry points (`execution.rs:19`, `:27`, `:78`), but **not** at `resume()` (`execution.rs:230`). Since `run_stackless_loop` calls `memory_check()` → `check_memory_budget_if_needed()` (`machine.rs:666`, `:469`), which enforces whenever `memory_budget_config.is_some()`, the budget is live on a resumed suspendable execution:

1. `set_execution_mode(Suspendable)`, no budget configured → `execute()` passes the guard and suspends.
2. `set_memory_budget_config(Some(..))` — no state or mode guard here (`machine.rs:430`); it captures a baseline immediately via `reset_memory_budget_state()` (`machine.rs:442`).
3. `resume()` → budget enforced against a baseline captured **outside** any execution, and only `reset_run_to_completion_state()` ever refreshes it.

If the resume lands on a different thread than step 2 — precisely the thread-migration scenario `docs/limits/memory_budget.md` says is unsafe and that this guard exists to prevent — the baseline belongs to an unrelated thread's counter. Result is either a spurious `MemoryBudgetExceeded` on a well-behaved policy or a silently disabled budget. Also reachable from `regorus_rvm_resume` (`bindings/ffi/src/rvm.rs:595`).

**Suggested fix** — guard `resume()` the same way:

```rust
pub fn resume(&mut self, resume_value: Option<Value>) -> Result<Value> {
    self.ensure_memory_budget_execution_mode()?;
    // Precondition is enforced below by returning `VmError::InvalidResumeState`
    ...
```

Optionally belt-and-braces: make `check_memory_budget_if_needed()` a no-op unless `execution_mode == RunToCompletion`, so enforcement can never key off a baseline that was not captured by `reset_run_to_completion_state()`.

Suggested test (none currently covers `resume()` with a budget):

```rust
// suspend first, then configure a budget, then resume
vm.set_execution_mode(ExecutionMode::Suspendable);
let _ = vm.execute();                       // suspends
vm.set_memory_budget_config(Some(MemoryBudgetConfig { limit: NonZeroU64::new(1 << 20).unwrap() }));
assert!(matches!(
    vm.resume(None),
    Err(VmError::MemoryBudgetUnsupportedInSuspendableExecution { .. })
));
```

src/rvm/vm/machine.rs
```diff
@@ -25,6 +27,9 @@ use super::execution_model::{
     BreakpointSet, ExecutionMode, ExecutionStack, ExecutionState, SuspendReason,
 };
 
+#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+const MEMORY_BUDGET_CHECK_STRIDE: u32 = 16;
```
anakrish: **[High] Stride-only sampling with no byte-delta trigger makes the budget strictly coarser than the global limit it takes precedence over**

*Finding from the testing/adversarial review; independently raised by the correctness review and the repo `code-review` skill.*

`check_memory_budget_if_needed()` (`machine.rs:469`) samples only once every `MEMORY_BUDGET_CHECK_STRIDE = 16` dispatch ticks and has **no** allocation-delta escape hatch. Compare the global path, `check_memory_limit_if_needed()` (`src/utils/limits/memory.rs:186-198`), which deliberately also trips on `pending_delta >= MEMORY_CHECK_DELTA_BYTES` (32 KiB) and on the allocator flush hint — precisely to catch bursts between strides.

Two consequences:

1. **Unbounded overshoot.** `memory_check()` is only reached at instruction boundaries (`dispatch.rs:28`, `execution.rs:136`, `:327`), so a single builtin — `json.unmarshal` of a large document, `numbers.range`, `sprintf`, string concat, a set union, or one comprehension producing a giant value — allocates its entire result with zero budget checks inside it. A 1 MiB budget does not bound peak RSS by 1 MiB; it can be exceeded by orders of magnitude. `docs/limits/memory_budget.md` notes that "enforcement can overshoot between checks" but gives no bound; for a control whose stated purpose is isolating individual evaluations, this should be documented as *unbounded within a single instruction*.
2. **The finer control is overridden by the coarser one.** `apply_memory_budget_precedence` (`machine.rs:507`) upgrades `MemoryLimitExceeded` into `MemoryBudgetExceeded`, so the budget takes priority over a check that is strictly more sensitive to bursts.

The stride also buys very little here: `current_thread_live_bytes()` is a single thread-local `Cell` read with no atomics and no allocator flush (`mimalloc/src/limits.rs:214-217`) — far cheaper than the global path it is being throttled more aggressively than.

**Suggested fix** — add a delta trigger mirroring the global path, so large bursts are caught regardless of stride:

```rust
#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
fn check_memory_budget_if_needed(&mut self) -> Result<()> {
    if self.memory_budget_config.is_none() {
        self.memory_budget_ticks = 0;
        return Ok(());
    }

    let next = self.memory_budget_ticks.saturating_add(1);
    let pending_delta = mimalloc::limits::thread_allocation_pending_delta().unsigned_abs() as u64;
    if next < MEMORY_BUDGET_CHECK_STRIDE && pending_delta < MEMORY_BUDGET_CHECK_DELTA_BYTES {
        self.memory_budget_ticks = next;
        return Ok(());
    }

    self.memory_budget_ticks = 0;
    self.check_memory_budget_now()
}
```

Given how cheap `current_thread_live_bytes()` is, simply dropping the stride and checking on every dispatch is also worth benchmarking — it would make the budget the tightest control in the system rather than the loosest.

No test currently measures overshoot. Suggested addition: a policy whose single builtin call allocates far more than the budget (e.g. `json.unmarshal` of a ~200 MB array with a 1 MiB budget), asserting the reported `usage` in `VmError::MemoryBudgetExceeded` stays within a stated multiple of `budget`.

tests/memory_limits.rs
```diff
@@ -283,3 +289,237 @@ fn add_data_conflict_is_atomic_on_allocator_build() {
...
+        .expect("compile policy for VM");
+    let program = Compiler::compile_from_policy(&compiled, &[entrypoint.as_ref()])
+        .expect("compile VM program");
+
+    let mut vm = RegoVM::new();
+    vm.load_program(program);
+    vm.set_data(engine.get_data()).expect("set data");
+    vm.set_input(Value::Undefined);
+    vm.set_memory_budget_config(Some(MemoryBudgetConfig {
+        limit: NonZeroU64::new(1).expect("non-zero budget"),
```
anakrish: **[High] Every budget-exhaustion test uses a 1-byte budget, so none of them prove the threshold is actually tracked**

*Finding from the testing/adversarial review; the same gap was flagged as missing coverage by the correctness review and the repo `code-review` skill.*

All six Rust exhaustion tests (`tests/memory_limits.rs:313`, `:340`, `:355`, `:370`, `:441`, `:494`), plus the FFI test (`bindings/ffi/src/rvm.rs:716`) and both C# tests (`RvmMemoryBudgetTests.cs:33,43`), configure `NonZeroU64::new(1)`. A 1-byte budget is exceeded by `initialize_execution_state()`'s own allocations (the `rule_cache` vec and register resize in `state.rs`) before the policy does anything meaningful. These tests would still pass if `check_memory_budget_now()` compared against a constant, sampled once, or measured memory unrelated to the execution. Nothing asserts the `usage` / `budget` / `pc` payload, so nothing pins `usage > budget` or that `budget` round-trips.

This matters because it interacts with two real behaviors elsewhere in the diff: the negative-delta clamp in `check_memory_budget_now()` (`machine.rs:490-492`, `saturating_sub` + `unwrap_or_default` floors usage at 0) and the stride-only sampling — both of which can silently disable enforcement without failing any current test.

Related: `vm_memory_budget_takes_precedence_over_global_limit` (`tests/memory_limits.rs:325`) sets budget = 1 **and** the global limit below current usage. Since `check_memory_budget_if_needed()` runs first in `memory_check()` (`machine.rs:666`), the global check is never reached — `apply_memory_budget_precedence()` (`machine.rs:507`) is never exercised by the test that is named after it.

**Suggested additions** (highest value first):

```rust
// 1. Threshold is real: same policy, generous budget succeeds; tight budget fails.
let generous = NonZeroU64::new(64 * 1024 * 1024).unwrap();
assert!(run_with_budget(generous).is_ok());
let tight = NonZeroU64::new(64 * 1024).unwrap();
assert!(matches!(run_with_budget(tight), Err(VmError::MemoryBudgetExceeded { .. })));

// 2. Payload is meaningful.
match run_with_budget(tight) {
    Err(VmError::MemoryBudgetExceeded { usage, budget, pc }) => {
        assert!(usage > budget, "usage {usage} must exceed budget {budget}");
        assert_eq!(budget, tight.get());
        assert!((pc as usize) < program.instructions.len());
    }
    other => panic!("expected budget error, got {other:?}"),
}

// 3. Exactly-at-limit must succeed — `usage > budget` (machine.rs:494) is untested.
// 4. u64::MAX budget: no overflow, no spurious failure.
// 5. Real precedence: global limit tight, budget generous => MemoryLimitExceeded
//    (proves the budget does not mask the global limit), and the reverse ordering
//    so `apply_memory_budget_precedence` is actually hit.
```

Also worth pinning: reuse of a VM after `MemoryBudgetExceeded` (larger budget then succeeds), and `set_memory_budget_config(None)` restoring prior behavior in Rust — currently only covered on the C# side (`RvmMemoryBudgetTests.cs:38`).

Minor, same file: `vm_memory_budgets_are_independent_across_threads` (`tests/memory_limits.rs:469`) is a good test — the `Barrier` forces genuine overlap — but the relaxed thread only asserts `Value::Array(_)`. Asserting the expected array length would prevent a truncated or short-circuited result from passing.

bindings/ffi/src/rvm.rs
```diff
@@ -489,10 +549,7 @@ pub extern "C" fn regorus_rvm_execute(vm: *mut RegorusRvm) -> RegorusResult {
             result.to_json_str()
```
anakrish: **[Medium] Result JSON serialization happens after the VM's final budget check, so it escapes the per-execution budget across FFI and C#**

*Finding from the `deep-review` skill (its sole material finding; six other candidates were disproved and dropped).*

`guard.execute()` performs its last budget check inside the VM — `check_memory_budget_now()` runs just before returning success (`src/rvm/vm/execution.rs:64`, `:113`, `:184`). `result.to_json_str()` then runs *outside* that boundary, on this same thread, and serializes the entire result value into a fresh `String`. For a policy that legitimately produces a large result under budget, the JSON encoding can be several times the size of the `Value` graph and is charged to nobody.

Same pattern in all three entry points: `rvm.rs:549`, `:568`, `:586` (and `:610` on the resume path). Every C# caller inherits this, since `Rvm.Execute*` goes through these exports — so the budget a C# host configures via `Rvm.SetMemoryBudgetConfig` does not cover the peak memory of the call it is meant to bound.

This is a smaller hole than in-instruction overshoot, but it is squarely inside the documented promise in `docs/limits/memory_budget.md` that the budget covers "additional current-thread live bytes above a fresh execution baseline" for the execution, and it is fully under the FFI layer's control.

**Suggested fix** — re-check after serialization while the baseline is still valid, e.g. expose a small helper on `RegoVM` and call it in each of the three (four, with resume) sites:

```rust
let output = || -> Result<String> {
    let vm = to_shared_ref(vm as *const RegorusRvm)?;
    let mut guard = vm.try_write()?;
    let result = guard.execute()?;
    let json = result.to_json_str()?;
    #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
    guard.check_memory_budget_now()?;   // needs pub(crate) -> pub, or a thin pub wrapper
    Ok(json)
}();
```

Note `check_memory_budget_now` is currently `pub(super)` (`machine.rs:487`), so this needs a narrow public wrapper — e.g. `pub fn verify_memory_budget(&self) -> Result<()>`, which would also give hosts a way to check the budget around their own post-execution work.

If serialization is instead considered deliberately out of scope (it is arguably "result handoff", like the excluded input/data loading), then state that explicitly in `docs/limits/memory_budget.md` alongside compilation and data/input/context loading, and in the C# xmldoc on `SetMemoryBudgetConfig` — right now the exclusion list does not mention it, so a host sizing a budget from that doc will underestimate peak memory.

bindings/csharp/Regorus/MemoryBudgetConfig.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// Configures the additional live-memory budget for one RVM execution.
+    /// </summary>
+    public readonly struct MemoryBudgetConfig
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MemoryBudgetConfig"/> struct.
+        /// </summary>
+        /// <param name="limitBytes">Maximum additional live bytes allowed during one execution.</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="limitBytes"/> is zero.</exception>
+        public MemoryBudgetConfig(ulong limitBytes)
```
anakrish, line 18: **[Medium] `default(MemoryBudgetConfig)` bypasses the non-zero constructor guard, and the fallback throws a different exception type**

*Finding from the API/FFI/bindings review; also raised by the repo `code-review` skill.*

`MemoryBudgetConfig` is a `readonly struct`, so C# guarantees a parameterless default construction that never runs this constructor. All of these produce `LimitBytes == 0` without any validation:

```csharp
var cfg = default(MemoryBudgetConfig);
var cfg2 = new MemoryBudgetConfig();          // legal for structs, does NOT call the ctor
var arr = new MemoryBudgetConfig[1];          // arr[0].LimitBytes == 0
class Host { MemoryBudgetConfig cfg; }        // field default
```

The `ToNative()` guard (line 35) catches these, so nothing invalid reaches the FFI — the safety property holds. But the failure surfaces as `InvalidOperationException` at `SetMemoryBudgetConfig()` call time (`Rvm.cs:151`), whereas the constructor throws `ArgumentOutOfRangeException` at construction time. A caller writing `catch (ArgumentOutOfRangeException)` around configuration will miss the default-struct case, and the error points at the wrong place — the API contract effectively has two different exception types for the same invalid value.

Note the Rust side does not have this hole: `MemoryBudgetConfig.limit` is `NonZeroU64`, so zero is unrepresentable. The C# mirror weakens that invariant.

**Suggested fix (preferred)** — make the invariant unrepresentable by storing the offset, so the default value is a valid 1-byte budget rather than an invalid 0:

```csharp
public readonly struct MemoryBudgetConfig
{
    private readonly ulong limitBytesMinusOne;

    public MemoryBudgetConfig(ulong limitBytes)
    {
        if (limitBytes == 0)
        {
            throw new ArgumentOutOfRangeException(nameof(limitBytes), "Memory budget must be non-zero.");
        }

        limitBytesMinusOne = limitBytes - 1;
    }

    public ulong LimitBytes => limitBytesMinusOne + 1;

    internal Regorus.Internal.RegorusMemoryBudgetConfig ToNative() =>
        new Regorus.Internal.RegorusMemoryBudgetConfig { limit_bytes = LimitBytes };
}
```

This mirrors `NonZeroU64` semantics exactly and removes the second guard entirely.

**Simpler alternatives**, if the offset trick is considered too clever:
1. Make it a `sealed class` — reference types have no silent default instance, so the constructor is the only path.
2. Keep the struct but make `ToNative()` throw `ArgumentOutOfRangeException(nameof(LimitBytes), ...)` for consistency with the constructor, and document that `default(MemoryBudgetConfig)` is invalid.

Either way, worth a test asserting the behavior of `default(MemoryBudgetConfig)` in `RvmMemoryBudgetTests.cs` — there is currently no coverage of it.

src/lib.rs
```diff
@@ -175,7 +175,7 @@ pub use utils::limits::PolicyLengthConfig;
 pub use utils::limits::{
     check_global_memory_limit, enforce_memory_limit, flush_thread_memory_counters,
     global_memory_limit, set_global_memory_limit, set_thread_flush_threshold_override,
-    thread_memory_flush_threshold,
+    thread_memory_flush_threshold, MemoryBudgetConfig,
```
anakrish, line 179: **[Low] New public export is feature-gated but has no `docsrs` annotation**

*Finding from the API/FFI/bindings review; also noted by the repo `code-review` skill.*

`MemoryBudgetConfig` is exported under `#[cfg(all(feature = "allocator-memory-limits", not(miri)))]` (`src/lib.rs:174-179`) but without a matching `#[cfg_attr(docsrs, doc(cfg(...)))]`. The repo convention is at `src/lib.rs:537-539`:

```rust
#[cfg(feature = "coverage")]
#[cfg_attr(docsrs, doc(cfg(feature = "coverage")))]
pub mod coverage {
```

Without it, docs.rs renders the type with no indication that a feature flag is required, so users hit a confusing "cannot find type" error. Same applies to `RegoVM::set_memory_budget_config` and `memory_budget_config` (`src/rvm/vm/machine.rs:430`, `:437`).

**Suggested fix:**

```rust
#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
#[cfg_attr(docsrs, doc(cfg(feature = "allocator-memory-limits")))]
pub use utils::limits::{
    check_global_memory_limit, enforce_memory_limit, flush_thread_memory_counters,
    global_memory_limit, set_global_memory_limit, set_thread_flush_threshold_override,
    thread_memory_flush_threshold, MemoryBudgetConfig,
};
```

Partly inherited — the sibling memory-limit exports in this same block already lack the annotation — so fixing the whole block together would be reasonable.

src/utils/limits/memory.rs
```diff
@@ -14,6 +15,13 @@ const MEMORY_CHECK_STRIDE: u32 = 16;
 // catch short bursts before they exceed typical entry budgets while still amortizing the atomic.
 const MEMORY_CHECK_DELTA_BYTES: u64 = 32 * 1024;
 
+/// Configuration for a fresh RVM memory budget.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct MemoryBudgetConfig {
```
anakrish, line 21: **[Low] `MemoryBudgetConfig` has a public field and is not `#[non_exhaustive]`, freezing its shape for semver**

*Finding from the repo `code-review` skill; also raised by the `deep-review` skill.*

`pub limit: NonZeroU64` (`src/utils/limits/memory.rs:19-21`) means every future field is a breaking change, and callers can construct it with struct literal syntax — which the tests already do (`tests/memory_limits.rs:312`). A memory budget is exactly the kind of config likely to grow options later (e.g. a soft-warning threshold, an overshoot tolerance, or opting serialization in/out — see the FFI serialization comment on this PR).

Cheapest to fix now, before the API ships, since it costs nothing today and is a breaking change later.

**Suggested fix:**

```rust
/// Configuration for a fresh RVM memory budget.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[non_exhaustive]
pub struct MemoryBudgetConfig {
    /// Maximum additional live bytes allowed during one execution.
    pub limit: NonZeroU64,
}

impl MemoryBudgetConfig {
    /// Create a budget allowing `limit` additional live bytes per execution.
    #[must_use]
    pub const fn new(limit: NonZeroU64) -> Self {
        Self { limit }
    }
}
```

Call sites then use `MemoryBudgetConfig::new(limit)` instead of the struct literal. Note `ExecutionTimerConfig` — the sibling this API is modeled on — should be checked for consistency; if it is also a bare public struct, either follow it deliberately or fix both.

bindings/ffi/src/rvm.rs
```diff
@@ -478,6 +491,53 @@ pub extern "C" fn regorus_rvm_set_execution_timer_config(
     })
 }
 
+/// Configure the per-VM memory budget for run-to-completion execution.
+#[cfg(feature = "allocator-memory-limits")]
+#[no_mangle]
+pub extern "C" fn regorus_rvm_set_memory_budget_config(
```
anakrish, line 531: **[Low] No FFI/C# getter for the memory budget config, unlike every sibling limit API**

*Finding from the API/FFI/bindings review.*

Rust exposes both `set_memory_budget_config` and `memory_budget_config()` (`src/rvm/vm/machine.rs:430`, `:437`), but only the setter is exported here. The global memory limit has both directions — `regorus_set_global_memory_limit` and `regorus_get_global_memory_limit` (`bindings/ffi/src/limits.rs:50`, `:62`) — so this is an inconsistency with the established pattern in the same feature area.

Practical impact: a C# host cannot read back what it configured, cannot assert configuration in its own tests, and cannot implement "restore previous budget" around a scoped override without tracking it separately.

**Suggested fix** — mirror the global-limit pair:

```rust
#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
#[no_mangle]
pub extern "C" fn regorus_rvm_get_memory_budget_config(vm: *mut RegorusRvm) -> RegorusResult {
    with_unwind_guard(|| {
        // return has_config + limit_bytes, following optional_u64_to_result's convention
    })
}

#[cfg(not(all(feature = "allocator-memory-limits", not(miri))))]
#[no_mangle]
pub extern "C" fn regorus_rvm_get_memory_budget_config(_vm: *mut RegorusRvm) -> RegorusResult {
    feature_disabled("regorus_rvm_get_memory_budget_config")
}
```

plus a `MemoryBudgetConfig?` property or `TryGetMemoryBudgetConfig` on C# `Rvm`.

If a getter is deliberately out of scope for this PR, consider making the Rust `memory_budget_config()` accessor `pub(crate)` for now so the public surface stays consistent across languages and the getter can be added everywhere at once.

bindings/ffi/src/rvm.rs
```diff
@@ -56,6 +56,19 @@ impl RegorusRvm {
     }
 }
 
+fn to_rvm_string_result(output: Result<String>) -> RegorusResult {
+    match output {
+        Ok(json) => RegorusResult::ok_string(json),
+        Err(err) => {
+            let status = match err.downcast_ref::<VmError>() {
+                Some(VmError::MemoryBudgetExceeded { .. }) => RegorusStatus::MemoryBudgetExceeded,
```
anakrish: **[Low] `MemoryBudgetUnsupportedInSuspendableExecution` degrades to generic `Error`, forcing callers to string-match**

*Finding from the API/FFI/bindings review; the brittleness it causes was independently flagged by the testing/adversarial review.*

This match maps only `MemoryBudgetExceeded` to a dedicated status; the new misconfiguration error falls into `_ => RegorusStatus::Error`. The consequence is visible in the PR's own test — `bindings/csharp/Regorus.Tests/RvmMemoryBudgetTests.cs:61` asserts:

```csharp
StringAssert.Contains(exception.Message, "not supported for suspendable execution");
```

That couples the test (and any host that wants to handle this case) to the exact `thiserror` string in `src/rvm/vm/errors.rs:36`, which is otherwise free to change. It is also a genuinely different class of failure from `MemoryBudgetExceeded`: this one is a programming error in configuration, detectable before execution, and always actionable by the caller — exactly what a distinct status is for.

**Suggested fix** — map it to the existing invalid-argument status:

```rust
let status = match err.downcast_ref::<VmError>() {
    Some(VmError::MemoryBudgetExceeded { .. }) => RegorusStatus::MemoryBudgetExceeded,
    Some(VmError::MemoryBudgetUnsupportedInSuspendableExecution { .. }) => {
        RegorusStatus::InvalidArgument
    }
    _ => RegorusStatus::Error,
};
```

and have the C# layer surface it as `ArgumentException`/`InvalidOperationException` based on the status rather than the message, so `RvmMemoryBudgetTests.cs:61` can assert on a type instead of a substring. (If `RegorusStatus::InvalidArgument` does not exist, appending a new variant is safe here for the same reason `MemoryBudgetExceeded` was — appended last, no discriminant shift.)

src/rvm/vm/state.rs
```diff
@@ -11,30 +11,45 @@ use super::machine::RegoVM;
...
     pub(super) fn reset_execution_state(&mut self) {
+        self.release_previous_execution_state();
+        self.initialize_execution_state();
+    }
+
+    /// Release values retained by the previous execution before capturing a new memory baseline.
+    pub(super) fn reset_run_to_completion_state(&mut self) {
+        self.release_previous_execution_state();
+        #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+        self.reset_memory_budget_state();
```
anakrish: **[Low] Baseline is captured before `initialize_execution_state()`, so per-execution setup allocations are charged to the policy's budget**

*Finding from the repo `code-review` skill; the same ordering detail was noted by the correctness review.*

`reset_memory_budget_state()` runs between `release_previous_execution_state()` and `initialize_execution_state()`. The ordering relative to *release* is the important and correct part — freeing the previous run's values before the baseline is exactly what prevents a prior execution from crediting the next one's budget, and `vm_memory_budget_does_not_receive_credit_from_previous_results` (`tests/memory_limits.rs:416`) covers it. That is the subtlest thing in this design and it is right.

The ordering relative to *initialize* is less clearly intended. `initialize_execution_state()` then allocates a fresh `rule_cache` (`state.rs:47`):

```rust
self.rule_cache = alloc::vec![(false, Value::Undefined); self.program.rule_infos.len()];
```

plus the `registers.resize(self.base_register_count, ..)`. Both land *after* the baseline, so VM setup proportional to program size is billed against the policy's budget. `docs/limits/memory_budget.md` describes the budget as covering execution work with compilation and data/input/context loading excluded; engine-internal setup scaling with `rule_infos.len()` sits awkwardly in that description. For a large program with a small budget, a non-trivial slice of the budget is consumed before the first instruction executes.

Two knock-on details worth deciding explicitly:
- `release_previous_execution_state()` uses `clear()` on `rule_cache`/`registers`/`cached_builtin_args`, which retains the buffers; the subsequent `self.rule_cache = vec![..]` drops the old buffer *after* the baseline, a small credit against the new budget.
- `register_window_pool` is never released, so pooled capacity grown during run 1 is free for run 2 — meaning identical runs are not billed identically.

**Suggested fix** — capture the baseline after setup, so the budget measures policy-attributable work:

```rust
pub(super) fn reset_run_to_completion_state(&mut self) {
    self.release_previous_execution_state();
    self.initialize_execution_state();
    #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
    self.reset_memory_budget_state();
}
```

This preserves the release-before-baseline property that the existing test pins. If the current ordering is deliberate (billing setup to the caller so budgets bound total per-execution footprint), that is defensible — but then say so in `docs/limits/memory_budget.md`, since the budget becomes program-size dependent and a budget tuned on a small program will not transfer to a large one.

bindings/csharp/Regorus/NativeMethods.cs
```diff
@@ -245,6 +245,12 @@ internal static unsafe partial class API
         /// </summary>
         [DllImport(LibraryName, EntryPoint = "regorus_rvm_set_execution_timer_config", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
         internal static extern RegorusResult regorus_rvm_set_execution_timer_config(RegorusRvm* vm, [MarshalAs(UnmanagedType.I1)] bool has_config, RegorusExecutionTimerConfig config);
+
+        /// <summary>
+        /// Set memory budget configuration.
+        /// </summary>
+        [DllImport(LibraryName, EntryPoint = "regorus_rvm_set_memory_budget_config", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
+        internal static extern RegorusResult regorus_rvm_set_memory_budget_config(RegorusRvm* vm, [MarshalAs(UnmanagedType.I1)] bool has_config, RegorusMemoryBudgetConfig config);
```
anakrish, line 253: **[Low] The new `DllImport` was inserted into the middle of another method's XML doc comment**

*Finding from the API/FFI/bindings review.*

The `regorus_rvm_set_memory_budget_config` declaration landed between the opening `/// <summary>` of `regorus_engine_add_policy` and the rest of that comment. The result (lines 249-258) is:

```csharp
/// <summary>
/// Set memory budget configuration.
/// </summary>
[DllImport(...)]
internal static extern RegorusResult regorus_rvm_set_memory_budget_config(...);
/// Add a policy.
/// The policy is parsed into AST.
/// See https://docs.rs/regorus/latest/regorus/struct.Engine.html#method.add_policy
/// </summary>
[DllImport(...)]
internal static extern RegorusResult regorus_engine_add_policy(...);
```

`regorus_engine_add_policy` now has a doc block with a closing `</summary>` and no opener. Cosmetic — it compiles, and the members are `internal` — but it will produce a malformed-XML warning under `GenerateDocumentationFile` and confuses doc tooling.

**Suggested fix** — move the new declaration below `regorus_engine_add_policy`'s comment block, or restore the missing opener:

```csharp
[DllImport(LibraryName, EntryPoint = "regorus_rvm_set_memory_budget_config", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
internal static extern RegorusResult regorus_rvm_set_memory_budget_config(RegorusRvm* vm, [MarshalAs(UnmanagedType.I1)] bool has_config, RegorusMemoryBudgetConfig config);

/// <summary>
/// Add a policy.
/// The policy is parsed into AST.
/// See https://docs.rs/regorus/latest/regorus/struct.Engine.html#method.add_policy
/// </summary>
[DllImport(LibraryName, EntryPoint = "regorus_engine_add_policy", ...)]
```

Also worth checking whether this file is generated or hand-maintained — if generated, the generator template needs the fix instead.

bindings/csharp/README.md
```diff
@@ -105,6 +105,29 @@ var result = vm.Execute();
...
+{
+    var result = vm.Execute();
+}
+catch (RegorusMemoryBudgetExceededException ex)
+{
+    Console.WriteLine(ex.Message);
+}
+```
+
+The budget is cooperative and may overshoot between VM checks. It is not supported in suspendable execution mode. `ClearMemoryBudgetConfig` restores the previous unlimited per-execution behavior. The process-wide limit exposed by `MemoryLimits` remains a separate safeguard.
```
anakrish: **[Low] `bindings/csharp/API.md` and both `CHANGELOG.md` files were not updated for the new public API**

*Finding from the API/FFI/bindings review.*

This README section is a good addition — the example is clear and the caveat on line 129 correctly states the cooperative/overshoot and suspendable-mode limitations. Three companion docs were missed:

- **`bindings/csharp/API.md`** — the canonical C# API listing; it contains no `MemoryBudget` references, so `Rvm.SetMemoryBudgetConfig`, `Rvm.ClearMemoryBudgetConfig`, `MemoryBudgetConfig`, and `RegorusMemoryBudgetExceededException` are absent from the documented surface.
- **`CHANGELOG.md`** — has an `## [Unreleased]` section (line 7) with no entry for `RegoVM::set_memory_budget_config` / `MemoryBudgetConfig` / the new `VmError` variants.
- **`bindings/ffi/CHANGELOG.md`** — likewise (line 7). This one matters most: a new `RegorusStatus` variant and a new exported symbol are exactly what downstream binding maintainers scan the FFI changelog for.

**Suggested additions:**

```markdown
<!-- CHANGELOG.md, under [Unreleased] / Added -->
- `RegoVM::set_memory_budget_config` and `MemoryBudgetConfig` for opt-in per-execution
  memory budgets on run-to-completion RVM executions (requires `allocator-memory-limits`).
  Adds `VmError::MemoryBudgetExceeded` and
  `VmError::MemoryBudgetUnsupportedInSuspendableExecution`.

<!-- bindings/ffi/CHANGELOG.md, under [Unreleased] / Added -->
- `regorus_rvm_set_memory_budget_config` export and `RegorusMemoryBudgetConfig` struct.
- `RegorusStatus::MemoryBudgetExceeded`, appended to the end of the enum
  (existing discriminants unchanged).
```

Also, `docs/limits/memory_budget.md` is not linked from any index — consistent with the existing `execution_time.md`, so acceptable, but linking both from the docs index would help discoverability. And per the separate comment on `docs`/scope: whichever way the FFI result-serialization question is resolved, the exclusion list in `docs/limits/memory_budget.md` should reflect it.

src/rvm/vm/machine.rs
```diff
@@ -402,6 +425,105 @@ impl RegoVM {
...
+
+    #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+    pub(super) fn check_memory_budget_now(&self) -> Result<()> {
+        let Some(config) = self.memory_budget_config else {
+            return Ok(());
+        };
+
+        let current = limits::current_thread_live_bytes();
+        let usage =
+            u64::try_from(current.saturating_sub(self.memory_budget_baseline)).unwrap_or_default();
```
anakrish: **[Medium] Clamping negative deltas to zero lets cross-thread frees silently buy free budget headroom**

*Finding from the correctness/memory-accounting review; independently raised by the testing/adversarial review and the repo `code-review` skill.*

### The mechanism

`memory_budget_baseline` is `i64` (`machine.rs:145`) and `current_thread_live_bytes()` returns an `i64` that can legitimately go negative. That is not an inference — the allocator layer documents it (`mimalloc/src/limits.rs`):

```rust
/// Current live bytes for this thread. May be negative if this thread freed
/// allocations that originated elsewhere.
pub allocated: i64,
```

Counters are incremented on alloc and decremented on dealloc *on the thread performing the operation*. When thread B frees memory that thread A allocated, B's counter drops without ever having risen, so `current` can fall below `baseline` mid-execution.

When that happens, `current.saturating_sub(baseline)` is negative, `u64::try_from(..)` fails, and `.unwrap_or_default()` yields **0**. The arithmetic itself is panic-free and correct — that part I verified. The issue is semantic: each sample is *instantaneous and independent*, with no memory of what was allocated earlier in the run. If an execution frees F bytes of foreign-allocated memory, `usage` reads 0 and the execution can then allocate F bytes entirely off the books before `usage` becomes non-zero again.

### Reachability — narrower than it first appears

I traced the paths, and several plausible ones are already handled correctly:

- Data, input, and the program stay alive for the whole run (the VM holds them), so they are never freed mid-execution.
- The previous run's values are freed in `release_previous_execution_state()` **before** the baseline is captured, so a thread migration *between* runs is absorbed into the baseline. That design is right, and `vm_memory_budget_does_not_receive_credit_from_previous_results` pins it.

The genuine path is global LRU eviction: `REGEX_CACHE` / `GLOB_CACHE` (`src/cache.rs:127,137`) are process-global `Mutex<LruCache>`, so an execution on thread B can evict a compiled `Regex`/`GlobMatcher` that thread A compiled, decrementing B's counter for memory it never allocated. `arc` is a default feature, so `Value`s are `Arc`-based and cross threads freely. This is attacker-influenceable in a multi-tenant server (flood distinct patterns to force evictions), though bounded by cache capacity (256 + 128 entries) rather than unbounded.

### Why it is still worth fixing

The fix is one line, and the failure mode is **silent fail-open**: the budget does not report a smaller number or fire late, it reads exactly `0` and stays disabled until usage climbs back — with no signal in production that enforcement lapsed. For an isolation control that is the worst failure mode.

More importantly it protects an *invariant* rather than just today's behavior. "Usage is monotone within an execution" keeps holding as the code evolves; without it, any future change that introduces cross-thread frees during execution silently weakens the budget and no test fails.

### Suggested fix

Preferred — ratchet the baseline down so usage can never decrease within a run:

```rust
#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
pub(super) fn check_memory_budget_now(&mut self) -> Result<()> {   // note: now &mut self
    let Some(config) = self.memory_budget_config else {
        return Ok(());
    };

    let current = limits::current_thread_live_bytes();
    // Foreign frees can push this thread's counter below the baseline. Ratchet the
    // baseline down so freeing memory allocated on another thread cannot buy budget.
    self.memory_budget_baseline = self.memory_budget_baseline.min(current);

    let usage =
        u64::try_from(current.saturating_sub(self.memory_budget_baseline)).unwrap_or_default();
    let budget = config.limit.get();
    if usage > budget {
        return Err(VmError::MemoryBudgetExceeded { usage, budget, pc: self.pc });
    }

    Ok(())
}
```

`apply_memory_budget_precedence` (`machine.rs:507`) takes `&self` and would need `&mut self` too; both callers already have a mutable borrow.

Alternative — track a per-execution high-water mark of `usage` instead of the instantaneous delta. Slightly more state but does not permanently lower the baseline, so it is marginally less strict. `ThreadAllocationStats` already exposes a `peak` field; snapshotting it at baseline and diffing may be cheaper than tracking it in the VM.

Either way, `docs/limits/memory_budget.md` should state the caveat: the budget measures **this thread's live-byte delta**, which is not the same as **this execution's allocations** once values cross threads. That distinction matters for anyone sizing a budget and is currently unstated.

### Priority

Lower than the stride/delta-trigger issue flagged separately on `machine.rs:31` — there, a single `json.unmarshal` overshoots by GB with no cross-thread setup at all. Both defeat the same property, but that one is trivially reachable and unbounded, whereas this needs a contrived setup for bounded headroom. Worth doing in this PR since it is one line and the code is fresh, but sequence it second. If scope is a concern, the acceptable minimum is the docs caveat above.

### Test gap

No current test can catch this: every exhaustion test uses a 1-byte budget on a single thread, so `usage` is non-zero immediately and the clamp never engages. `vm_memory_budgets_are_independent_across_threads` builds *and* executes each VM on its own thread, so there are no cross-thread frees. A regression test would construct the VM and populate the caches on thread A, execute on thread B, and assert the budget still fires.

bindings/ffi/src/rvm.rs
```diff
@@ -478,6 +491,53 @@ pub extern "C" fn regorus_rvm_set_execution_timer_config(
     })
 }
 
+/// Configure the per-VM memory budget for run-to-completion execution.
```
anakrish, line 528: **[Medium] FFI memory-budget gates do not match the core `not(miri)` gate**

*Finding from the API/bindings review; independently flagged by the repo `code-review` skill.*

The core API and implementation are gated with:

```rust
#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
```

For example, `MemoryBudgetConfig` is not exported from `src/lib.rs` under Miri and `RegoVM::set_memory_budget_config` is absent from `machine.rs` in that configuration. But the FFI setter and FFI test module use only:

```rust
#[cfg(feature = "allocator-memory-limits")]
// bindings/ffi/src/rvm.rs:494

#[cfg(all(test, feature = "allocator-memory-limits")]
// bindings/ffi/src/rvm.rs:638
```

With `allocator-memory-limits` enabled under Miri, the FFI crate therefore tries to compile code that references `regorus::MemoryBudgetConfig` / `guard.set_memory_budget_config(...)` and the budget test imports, even though the core crate deliberately does not expose those items under Miri. This is currently latent because `.github/workflows/miri.yml` runs Miri only for the core crate, but it will surface as soon as FFI Miri coverage or a downstream all-workspace Miri build is enabled.

This is also a maintenance hazard: the two layers now disagree about whether the feature is available, so future CI changes can turn a currently hidden configuration bug into a build break.

**Suggested fix:** mirror the core predicate in both FFI locations and retain the existing feature-disabled symbols under the inverse predicate:

```rust
#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
#[no_mangle]
pub extern "C" fn regorus_rvm_set_memory_budget_config(...) -> RegorusResult {
    // existing implementation
}

#[cfg(not(all(feature = "allocator-memory-limits", not(miri))))]
#[no_mangle]
pub extern "C" fn regorus_rvm_set_memory_budget_config(...) -> RegorusResult {
    feature_disabled("regorus_rvm_set_memory_budget_config")
}

#[cfg(all(test, feature = "allocator-memory-limits", not(miri)))]
mod tests {
    // existing budget tests
}
```

The exported disabled stub is important: it preserves the symbol for bindings built against configurations where allocator memory limits are unavailable. Add a CI compile job for the FFI crate under Miri (or at least `cargo check -p regorus-ffi --features allocator-memory-limits` with the relevant Miri cfg) so this cannot regress.

This is a configuration/build correctness issue rather than a runtime-budget issue; it does not affect the default non-Miri build.

anakrish:
@maksym-mishchenko Thanks for doing this very useful feature! Overall looks good to me. Copilot reviews found some interesting cases that are worth addressing.

src/rvm/vm/execution.rs
maksym-mishchenko: Fixed in fbf805c. `resume()` now applies the execution-mode guard before entering the VM loop, and budgeted suspendable execution is rejected with the typed `MemoryBudgetUnsupportedInSuspendableExecution` status and managed exception.

src/rvm/vm/machine.rs
maksym-mishchenko: Fixed in fbf805c. The 16-check stride was removed; the budget is now sampled at every VM memory checkpoint.

tests/memory_limits.rs
maksym-mishchenko: Fixed in fbf805c. The 1-byte cases were replaced with realistic thresholds that allow execution setup to complete and exercise budget exhaustion during the tested workload.

bindings/ffi/src/rvm.rs
maksym-mishchenko: Fixed in fbf805c. FFI now serializes into a provisional native string, rechecks the execution budget after serialization, and frees that provisional output before returning the budget-exceeded status on failure.

bindings/csharp/Regorus/MemoryBudgetConfig.cs
maksym-mishchenko: Fixed in fbf805c. `SetMemoryBudgetConfig` validates `LimitBytes`, so `default(MemoryBudgetConfig)` now throws `ArgumentOutOfRangeException`, consistent with constructor validation.

src/lib.rs
maksym-mishchenko: Fixed in fbf805c. Added matching docs.rs `doc(cfg(...))` annotations for the feature-gated memory-budget export and RVM APIs.

src/utils/limits/memory.rs
maksym-mishchenko: Intentionally not changed. Sibling configuration types use public fields, and retaining struct-literal construction preserves the established Rust API style rather than introducing a one-off pattern for this config.

bindings/ffi/src/rvm.rs
maksym-mishchenko: Intentionally not added. The closest per-RVM configuration precedent, the timer API, is setter-only; unlike the process-global limit, no current FFI or managed caller needs budget readback, so the surface remains minimal.

bindings/ffi/src/rvm.rs
maksym-mishchenko: Fixed in fbf805c. Added the typed `MemoryBudgetUnsupportedInSuspendableExecution` FFI status and corresponding managed exception, so callers no longer need to string-match the generic error.

src/rvm/vm/state.rs
maksym-mishchenko: The ordering is intentional: execution-state setup must count toward the per-execution budget, so the baseline remains before initialization. The memory-budget documentation now states this explicitly.

bindings/csharp/Regorus/NativeMethods.cs
maksym-mishchenko: Fixed in fbf805c. The memory-budget declaration now has its own well-formed XML documentation block, and the `regorus_engine_add_policy` documentation is contiguous again.

bindings/csharp/README.md
maksym-mishchenko: Fixed in fbf805c. Updated the root and FFI changelogs and documented the new memory-budget configuration, methods, and managed exceptions in `bindings/csharp/API.md`.

src/rvm/vm/machine.rs
maksym-mishchenko: Fixed in fbf805c. The baseline now ratchets downward when sampled thread live bytes decrease, preventing already-observed foreign frees from creating budget headroom. The remaining between-sample netting limitation is documented because fully eliminating it requires allocator ownership tracking or scoped heaps.

bindings/ffi/src/rvm.rs
maksym-mishchenko: Fixed in fbf805c. All FFI allocator-memory-limit exports and tests now use the same `not(miri)` gating as the core API.

maksym-mishchenko:
> Maksym (@maksym-mishchenko) Thanks for doing this very useful feature! Overall looks good to me. Copilot reviews found some interesting cases that are worth addressing.

Hi @anakrish, thanks for the thorough review. I addressed comments, I kept the two API suggestions unchanged for the reasons explained in their threads. Could you please take another look when you have time?

benches/rvm_benchmark.rs
```diff
@@ -374,6 +389,11 @@ fn configure_limits(vm: &mut RegoVM, limits: bool) {
         vm.set_execution_timer_config(None);
         vm.set_max_instructions(usize::MAX);
     }
+
+    #[cfg(feature = "allocator-memory-limits")]
+    vm.set_memory_budget_config(config.memory_budget.then(|| MemoryBudgetConfig {
```
anakrish, line 397: **Low:** These benchmark imports/calls use `#[cfg(feature = "allocator-memory-limits")]`, but the core `MemoryBudgetConfig` export and `RegoVM::set_memory_budget_config` are gated by `all(feature = "allocator-memory-limits", not(miri))`. A bench build selected under Miri can therefore fail to compile.

Suggested change:
```rust
#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
use std::num::NonZeroU64;

#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
use regorus::MemoryBudgetConfig;

#[cfg(all(feature = "allocator-memory-limits", not(miri)))]
vm.set_memory_budget_config(config.memory_budget.then(|| MemoryBudgetConfig {
    limit: NonZeroU64::new(MEMORY_LIMIT_BYTES).expect("non-zero memory budget"),
}));
```

Please use the same predicate for both imports and the configuration block.

bindings/ffi/src/rvm.rs
```diff
@@ -56,6 +58,50 @@ impl RegorusRvm {
...
+{
+    let output = || -> Result<RegorusResult> {
+        let vm = to_shared_ref(vm as *const RegorusRvm)?;
+        let mut guard = vm.try_write()?;
+        let value = execute(&mut guard)?;
+        let json = value.to_json_str()?;
+        let result = RegorusResult::ok_string(json);
+
+        #[cfg(all(feature = "allocator-memory-limits", not(miri)))]
+        if let Err(err) = guard.check_memory_budget() {
```
anakrish: **Medium:** If this post-serialization budget check fails, the FFI returns `MemoryBudgetExceeded` after dropping only the provisional JSON result, but the VM has already stored the value as `ExecutionState::Completed { result }`. A caller can then call `regorus_rvm_get_execution_state()` and observe/re-serialize the oversized completed result despite the reported failure.

Suggested shape:
```rust
if let Err(err) = guard.check_memory_budget() {
    regorus_result_drop(result);
    guard.mark_execution_error(err.clone()); // clear retained result/state
    return Err(err.into());
}
```

Alternatively, move the final serialization check into a core-owned completion helper that transitions the VM to `ExecutionState::Error` and releases the retained result before returning. Please add a regression asserting the state is `Error` after this failure.

src/rvm/vm/execution.rs
```diff
@@ -163,10 +174,17 @@ impl RegoVM {
...
-        self.reset_execution_state();
+        self.reset_run_to_completion_state();
         self.reset_execution_timer_state();
         self.execution_state = ExecutionState::Running;
-        match self.jump_to(0_u32) {
+        let result = self
+            .jump_to(0_u32)
+            .map_err(|err| self.apply_memory_budget_precedence(err))
+            .and_then(|value| {
+                self.check_memory_budget()?;
```
anakrish: **Medium:** When this final budget check fails, the `Err` arm records `ExecutionState::Error` but leaves the failed execution allocations (`registers`, `rule_cache`, `evaluated`, and pooled values) resident until the next execution or VM drop. The named/indexed entry-point paths and the FFI post-marshalling failure do not clean them up at all. Reusing a VM after repeated oversized failures can therefore accumulate roughly one failed result/state per cycle.

Suggested shape:
```rust
fn fail_execution(&mut self, err: VmError) -> VmError {
    self.release_previous_execution_state();
    self.execution_state = ExecutionState::Error { error: err.clone() };
    err
}
```

Use this helper in the run-to-completion `Err` arm, the named/indexed entry-point budget-error paths, and the FFI post-marshalling failure path. Preserve the error value while releasing the result/register/cache state, and add a regression that repeats budget failures on a reused VM and verifies memory/state are reset.

anakrish:
> > Maksym (Maksym (@maksym-mishchenko)) Thanks for doing this very useful feature! Overall looks good to me. Copilot reviews found some interesting cases that are worth addressing.
> 
> Hi Anand Krishnamoorthi (@anakrish), thanks for the thorough review. I addressed comments, I kept the two API suggestions unchanged for the reasons explained in their threads. Could you please take another look when you have time?

The rationale for both sounds good. Review rerun found 3 more comments worth addressing. Then it should be good to go.

kusha:
@maksym-mishchenko I see the following drawbacks in this implementation:


- The budget is armed at `execute*()` time, not at `set_memory_budget_config()` time (`state.rs:19-22` → `machine.rs:442-449`). `set_memory_budget_config` sets `baseline = 0; active = false`; the real baseline is captured on the next execution. So it is a *per-execution* budget, not a budget window a host can open and close.
- `RegoVM::set_data` / `set_input` / `set_context` / `load_program` all run before the baseline and are uncharged (stated in `docs/limits/memory_budget.md`). Via FFI, a host can set a budget, load a 500 MB data blob successfully, and only then have evaluation-on-top bounded. If the goal is per-request isolation, the hostile-input case is the one not covered.

Why haven't you pinned thread level baseline and used every allocation (delta gated) for comparison (just like a global limit)? Why have you chosen execute() calls?

Consider whether the intended primitive is a *scope* (`begin(limit)` / `end()` pinning a thread baseline, consulted by every allocating entry point — data load, compile, execute, serialize) rather than a per-execution budget. `set_memory_budget_config` is about to become public API with per-execution semantics baked in.

@anakrish any thoughts on the above ^? 

---

Apart from the design:

**Follow-ups on the fixes from the previous review round**

- **Amortization was removed rather than refined** (`machine.rs:673`, `dispatch.rs:28`). Comment `3816390968` argued stride-16 sampling was too *coarse* and asked for a byte-delta trigger; the fix removed the stride entirely. `check_memory_budget()` now runs on every instruction, doing a `THREAD_COUNTERS.with(...)` TLS access per instruction whenever a budget is configured — while the global limit it takes precedence over is still amortized (`MEMORY_CHECK_STRIDE = 16` + 32 KiB delta gate). The bench harness gained a `regular_memory_budget` config; could the numbers be posted?
- **The `.min()` ratchet can make the budget stricter than configured** (`machine.rs:498`). `memory_budget_baseline = memory_budget_baseline.min(current)` never recovers, so any free of pre-baseline memory observed on the execution thread permanently lowers the baseline. Fail-safe direction, but it produces non-deterministic `MemoryBudgetExceeded` below the configured limit. `release_previous_execution_state()` before baseline capture mitigates the main in-engine case. The doc says only "does not grant credit" — it should also state the budget can become *stricter* than configured.

**New**

- **`memory_budget_active` is never cleared** (`machine.rs:448`; only reset at `machine.rs:431` and on the next execution). After a successful `execute()` the window stays live, so a Rust host calling the public `check_memory_budget()` later charges unrelated work to the last execution. Given it is `pub` and the doc invites bindings to call it post-serialization, either consume the window on read or document it as valid only immediately after execute, on the same thread.
- **Binding parity** — only C# is wired up. C++ (`bindings/cpp/regorus.hpp`), Go, Java, Python, Ruby, and WASM get the new `RegorusStatus` variants but no budget API. Fine as an increment, but worth an explicit note in the PR or a tracking issue.
- **Doc overstatement** — `docs/limits/memory_budget.md` says the check covers "native string marshaling" for C#. The managed-string copy happens after the FFI call returns and is on the managed heap, not the mimalloc heap, so it is not covered.

anakrish:
> * `RegoVM::set_data` / `set_input` / `set_context` / `load_program` all run before the baseline and are uncharged (stated in `docs/limits/memory_budget.md`). Via FFI, a host can set a budget, load a 500 MB data blob successfully, and only then have evaluation-on-top bounded. If the goal is per-request isolation, the hostile-input case is the one not covered.

@kusha 
I can see rationale for both sides. If the data/input are not included in the evaluation budget then, as you showed, really large objects can escape the memory check.
On the other hand, it is easier for the caller to reason about/estimate the size of input and data (say from the length of the byte streams from which they are originally deserialized) and perform checks whereas it is harder for the caller to reason about a policy evaluation's memory consumption. This PR addresses the latter. 

However, I do see that it would be nice to also limit input, data memory consumption.

Some challenges around implementation:
- Setting input, data and evaluation is not an atomic operation. One thread could set input/data and another could perform the evaluation. Per thread budgets don't play nicely with this.
- If input/data exist as C# objects as opposed to a byte stream, then they must anyways be serialized on the C# size..ie the 500MB limit would be exceeded even before reaching rust.

Some ideas
- Should we have individual limits around data/object sizes?
- Or for evaluation, should the base line start at earlier or set data/input and end at the end of evaluation? Ideally in case of FFI after the result has been serialized.

anakrish:
> Why haven't you pinned thread level baseline and used every allocation (delta gated) for comparison (just like a global limit)? Why have you chosen execute() calls?

Maybe this would be the simplest approach. Have a thread level limit on alive memory?
But if one thread is the producer of programs and input/data and other threads are merely users, then it may not bounded correctly.

anakrish:
> So it is a per-execution budget, not a budget window a host can open and close.

Yes. That was also my impression from the PR description and the implementation. An execution-level budget is the clearest semantic: it bounds the additional working memory used during one RVM execution and intentionally excludes objects such as input, data, and the program that may have been created before—and may outlive—that execution.

However, I do agree that large data/input may go undetected by this budget, even if they are eventually subject to configured global memory limit.

We should meet to discuss your scenario and see if the design can be generalized and the semantics made more clear/easy to reason about.

maksym-mishchenko:
@kusha @anakrish. I made another pass based on this discussion.

The existing `execute*` APIs still have a fresh budget for each execution. I added an opt-in path that starts before evaluation-specific JSON data is parsed and set, then stays active through execution and native result serialization.

I avoided a `begin/end` API across FFI calls because the accounting is thread-local, and separate calls may run on different OS threads. Rust uses a same-thread scope; FFI and C# do everything in one call.

Program loading and compilation are still excluded. I also left input and context out for now. Is covering data enough for the fetch scenario, or do you think input should be included before we approve the API?

The cleanup and documentation issues are fixed as well. The benchmark showed no measurable difference: 859.82 ns without the budget and 847.59 ns with it, with overlapping intervals.

anakrish:
@maksym-mishchenko `data` is typically the data that doesn't share that often across individual policy evaluation. `input` is the actual input for a policy evaluation. IMO input would also need to be guarded.

anakrish:
I found a performance issue in the new memory-budget enforcement path: `memory_check()` is invoked twice per VM instruction in run-to-completion execution. `src/rvm/vm/execution.rs` calls `self.memory_check()` before dispatch, and `src/rvm/vm/dispatch.rs` calls it again at the start of `execute_instruction()`. The same duplicate exists on the suspendable dispatch path.

With allocator limits enabled, each call can sample the budget and run the throttled global-limit check, so this adds avoidable hot-path overhead and makes checkpoint cadence effectively 2x what the execution loop suggests.

Suggested fix: retain the check at one dispatch boundary (preferably the outer execution loop, unless a specific instruction-level invariant requires the inner check), then add a test-only check counter or benchmark assertion to verify one check per dispatched instruction. Please also compare the benchmark with one vs. two checks so the overhead is measurable.

anakrish:
Documentation suggestion — cooperative enforcement and overshoot:

Please make the hard-cap distinction especially explicit in `docs/limits/memory_budget.md` and the C# README/API docs. The budget is checked at VM checkpoints and after native result serialization; a single instruction, builtin, `to_json_str()`, or `CString` allocation can temporarily exceed the configured limit before the next check. This is a cooperative observed-live-bytes budget, not an allocation-time peak-memory cap. A short example of this behavior would help callers choose an appropriate headroom.

anakrish:
Documentation suggestion — VM reuse and warm pools:

Please document that budget outcomes can depend on prior executions when the same `RegoVM` is reused. Execution state is reset and the baseline is fresh, but retained VM pools/capacities are already live outside the new baseline, so an identical policy/input may have different allocation behavior on a warm VM than on a fresh VM. This is an important qualification to the phrase “fresh per-execution budget.”

anakrish:
Documentation suggestion — baseline ratcheting and the `usage` diagnostic:

Please clarify that the baseline is lowered when sampled same-thread live bytes fall below the prior baseline, and that this lost headroom is not restored. As a result, the effective budget can become stricter after unrelated or legitimate same-thread frees, and `VmError::MemoryBudgetExceeded { usage, .. }` should not be interpreted as exact memory owned by the policy evaluation or as a net-allocation measurement.

anakrish:
Documentation suggestion — host callbacks and cross-thread frees:

Please call out that accounting is based on the current thread’s live-byte counter, not allocation ownership. Allocations/frees performed by synchronous custom or host builtins on the execution thread are therefore observed as part of the budget; objects allocated on one thread and freed on another can temporarily skew per-thread/global observations until counters are published. This would help embedding applications avoid assuming strict query-owned attribution or thread-independent accounting.

benches/rvm_benchmark.rs
maksym-mishchenko: Fixed. The benchmark import and call sites now consistently use #[cfg(all(feature = "allocator-memory-limits", not(miri)))], including NonZeroU64 and set_memory_budget_config. The synthetic cfg(miri) benchmark build also passes.

bindings/ffi/src/rvm.rs
maksym-mishchenko: Fixed. Native execution now serializes through the RegoVM execute_*_to_c_string_for_ffi helpers. If JSON/CString creation or the post-serialization budget check fails, execute_to_c_string_for_ffi_with routes the error through fail_run_to_completion, which releases retained state and stores ExecutionState::Error. The FFI regression covers main, named, and indexed entry points, verifies MemoryBudgetExceeded with a null output and Error state, and then verifies successful VM reuse after clearing the budget.

src/rvm/vm/execution.rs
maksym-mishchenko: Fixed. Every run-to-completion error arm now funnels through fail_run_to_completion. It clears evaluated values, execution/register stacks, pooled register contents, rule and builtin caches, registers, cached builtin arguments, and the active budget lifecycle before storing ExecutionState::Error. failed_run_to_completion_releases_retained_values verifies the cleanup, and the repeated-execution tests verify a reused VM starts with fresh accounting.

anakrish (approved):
@maksym-mishchenko LGTM. Thanks for the contribution! We can revisit accounting of `data`, `input`, suspend/resume later based on actual usecases.

